Issue #22, opened 2020-10-29
Memory Game


tmansuroglu:
Create memory based image matching game.Add tests. Game should be working on different page. Add description and guidelines for the game.

PR #25, opened 2020-11-06
Memory game initial card state generation

tmansuroglu:
These are parts of bigger code. To prevent test error, parts related to other PRs are commented out. Screenshot not added because it is not meaningful without other components

#22


src/components/memoryCardBackSideImages.jsx
```diff
@@ -0,0 +1,29 @@
...
+import deforestation from "../images/memoryCardBack/deforestation.svg";
+import globalWarming from "../images/memoryCardBack/globalWarming.svg";
+import greenhouseEffect from "../images/memoryCardBack/greenhouseEffect.svg";
+import littering from "../images/memoryCardBack/littering.svg";
+import meltingIceCaps from "../images/memoryCardBack/meltingIceCaps.svg";
+import ozoneHole from "../images/memoryCardBack/ozoneHole.svg";
+import recycle from "../images/memoryCardBack/recycle.svg";
+import waterPollution from "../images/memoryCardBack/waterPollution.svg";
+
+const cardBacksideImages = [
```
louisrli: CARD_BACKSIDE_IMAGES

src/components/memoryCardBackSides.jsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import cardBacksideImages from "./memoryCardBackSideImages";
+// import "./MemoryGame/index.css";
+
+// This function will be used in "MemoryGame" component
+const cardBackSidesArr = (cardsData) => {
+  // cardData and cardBacksideImages ordered in same way.
+  // For example: cardData[0] holds data for cardBacksideImages[0]
```
louisrli: generally we want to avoid having "parallel lists" like this. though the comment is good, still that type of code (you will find with experience, i guess i encourage you to try it in another project to feel the pain of it) is hard to maintain and prone to error

what i would do:
* install the "invariant" package in npm
* `invariant(cardData.length === cardBacksideImages.length)`
* `const cardDataAndBacksides = []`. push [cardData[0], cardBacksideImages[0]] to the list...so it's a list of lists.
* you can now safely loop through cardDataAndBacksides such that its paired correctly; bc of your invariant you are sure that things are paired together correctly

src/components/memoryCardBackSides.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return cardBacksideImages.map((image, index) => {
+    const cardData = cardsData[index];
+    return (
+      // matching cards should have same id
+      <div key={cardData.description} id={index}>
```
louisrli: avoid using index as key

src/containers/MemoryGame/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+    [isMeltingIceCapsCloneFlipped, setIsMeltingIceCapsCloneFlipped],
+    [isOzoneHoleCloneFlipped, setIsOzoneHoleCloneFlipped],
+    [isRecycleCloneFlipped, setIsRecycleCloneFlipped],
+    [isWaterPollutionCloneFlipped, setIsWaterPollutionCloneFlipped],
+  ];
+
+  return (
+    <div className="memoryGameContainer">
+      <Navbar />
+      {/* <MemoryGameComponent cardStates={cardStates} /> */}
```
louisrli: no commented code in PRs as a general rule

src/containers/MemoryGame/index.jsx
```diff
@@ -0,0 +1,97 @@
+import React, { useState } from "react";
+// import MemoryGameComponent from "../../components/MemoryGame";
+import Navbar from "../../components/Navbar";
+import "./index.css";
+
+const MemoryGame = () => {
+  // Individual card states are being used
+  // Because stacking all card states inside 1 states causes unwanted side effects
+  const [isAcidRainFlipped, setIsAcidRainFlipped] = useState(false);
```
louisrli: as a general rule of thumb as a programmer you should always think this way: "if i have 5 things, how hard would it be for me to make it 10000 things?" the answer in this code is, very hard, because you need to add 9995 more states.

this code needs to be written such that it works for an arbitrary number N cards (and it is possible). so when you rewrite this code, think very very carefully about this rule of thumb i've given (and most code you write in your life from now on, there are exceptions where sometimes its faster to write things for the case of 2 or 3 but never for the case of 10-15 like this).

to give you a start, what you probably want to do is:
```
const cardFlippedState = React.useState([new Array(cardBackImages.length).fill(false)]);

const flipCard = (index: number) => {
   // mutate the state here
}

```
you can also consider making your state an object mapping image names to true/false rather than using indexes (cleaner and easier to debug IMO)

how i personally would do it, in the most extensible / flexible way to write code possible:
* cardSpecs.jsx. contains data about cards, contains an object `const CARD_SPECS` that has keys "string" and maps to "objects". objects includes backImage, but its extensible to include more data later if i need to add more data about cards (say "hints" or something)
* cardFlippedState would be an object that maps strings to booleans. and i would make it by looping through Object.keys(CARD_SPECS) and making it false by default
* flipCard would take a cardKey name, not an index (working with indexes is very unreadable in general)
* CARD_SPECS again is the source of truth and can be used to render all things on the page

but i want to emphasize this: never after today should you write code that has this many states that do the same thing. it's not one of those things where it's like "oh sometimes there's exceptions."

src/containers/MemoryGame/index.jsx
tmansuroglu: Line 92 will be used when it is merged with other PRs. Should I still delete it?

src/components/memoryCardBackSideImages.jsx
tmansuroglu: PTAL

src/components/memoryCardBackSides.jsx
tmansuroglu: `key `is already `cardData.description`.

If you meant `id={index}`, the reason behind it is that matching cards should have same id. Since there should be 2 of each card, after creating array with this function, i duplicate same array and put old and new array together in one array. I could do the same by creating ids by hand. But this way felt more logical. Should I still remove it?

src/components/memoryCardBackSides.jsx
tmansuroglu: PTAL

src/components/memoryCardBackSides.jsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import cardBacksideImages from "./memoryCardBackSideImages";
```
majdajroudi: I think the convention here is CardBackSideImages or CardBacksideImages, I'm not sure but the first letter for a component should be in caps

src/components/memoryCardBackSides.jsx
louisrli: sorry i misread but you should also avoid using index as id. first, the id would be just "0", "1", "2", which is unreadable; IDs are global. second, ID shouldn't be used anyway except in very very rare situations. it also doesn't make sense that matching cards should have the same ID because IDs are globally unique, you can only have one div with that id

tmansuroglu: PTAL

src/containers/MemoryGame/index.jsx
tmansuroglu: PTAL

src/containers/MemoryGame/index.jsx
tmansuroglu: Now code is reusable.
` const initialStates = {};
cardsArr.forEach((card) => {
    initialStates[card.key] = false;
  });
const [cardStates, setCardStates] = useState(initialStates);`

State gets manipulated with `setCardStates(prevState=>{...prevState,key:state})`

Also moved code to a more appropriate file. In addition I discovered the reason behind my first issue. Before using 20 states, I was trying to write something that is reusable(1 state to control all cards). After trying for 4.5 hours I couldn't solve problem of inaccurate flips and twitches and decided to create individual states. Today I was able to recreate same issue by passing state to children. When I moved `cardStates` where it is being used issue resolved. I don't know the exact reason behind it. But i plan to find out soon.

PTAL

src/components/memoryCardBackSideImages.jsx
```diff
@@ -0,0 +1,29 @@
+import acidRain from "../images/memoryCardBack/acidRain.svg";
+import airPollution from "../images/memoryCardBack/airPollution.svg";
+import alternateEnergy from "../images/memoryCardBack/alternateEnergy.svg";
+import chemicalWaste from "../images/memoryCardBack/chemicalWaste.svg";
```
louisrli: this file should be under src/components/memorygame/BackSideImages or something like this, its polluting the components space right now

also if you are going to hold info for the cards other than the backside images i would very much highly recommend doing the object approach that i said earlier (aka if you have a frontside or other facts associated with a card, i would not accept code that keeps them separately and correlates them using the array indexes)

the reason is that its extremely bug prone and you're relying on a human not to swap the order accidentally in one of the arrays. who is to say youre not going to make alternateEnergy 3rd instead of 4th accidentally in the front, but 4th instead of 3rd in the array for the back?

i see its in alphabetical order right now but in the end youre relying on a human to keep your things synchronized. again, think about if you had 10,000 cards

src/components/shuffle.jsx
```diff
@@ -0,0 +1,12 @@
+// Shuffles cards
```
louisrli: it should be clarified that this is not an in-place shuffle in this comment, which would be the first thing a reader wants to know when they see a fn called shuffle. not whether or not
 the fn called shuffle "Shuffles cards" (comments should be used to clarify non-obvious things or mark headers in longer files)

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
...
+    const { description } = dataAndImage[0];
+    const image = dataAndImage[1];
+    const isClone = cardBackSides.find(
+      (prevDescription) => prevDescription === description
+    )
+      ? true
+      : "";
+    cardBackSides.push(description);
+    return (
+      <div key={isClone ? `Clone of ${description}` : description}>
```
louisrli: if you are returning jsx here, then this whole thing should be a component, not a function

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import * as invariant from "invariant";
+import CARD_BACKSIDE_IMAGES from "./memoryCardBackSideImages";
+import shuffle from "./shuffle";
+
+// This function will be used in "MemoryGame" component
+const cardBackSidesArr = (cardDataArr) => {
```
louisrli: this is called Arr but it's a function

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
...
+  invariant(
+    cardDataArr.length === CARD_BACKSIDE_IMAGES.length,
+    "'CARD_BACKSIDE_IMAGES' and 'cardData' are not matching"
+  );
+  const cardDataAndBacksides = cardDataArr.map((cardData, index) => {
+    const cardBacksideImage = CARD_BACKSIDE_IMAGES[index];
+    return [cardData, cardBacksideImage];
+  });
+  const cardDataAndBacksidesWithClones = [
```
louisrli: needs comments why there are clones. imagine you are me. why are there clones?

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import * as invariant from "invariant";
+import CARD_BACKSIDE_IMAGES from "./memoryCardBackSideImages";
+import shuffle from "./shuffle";
+
+// This function will be used in "MemoryGame" component
+const cardBackSidesArr = (cardDataArr) => {
+  invariant(
+    cardDataArr.length === CARD_BACKSIDE_IMAGES.length,
+    "'CARD_BACKSIDE_IMAGES' and 'cardData' are not matching"
```
louisrli: are not matching length

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+  const cardDataAndBacksides = cardDataArr.map((cardData, index) => {
+    const cardBacksideImage = CARD_BACKSIDE_IMAGES[index];
+    return [cardData, cardBacksideImage];
+  });
+  const cardDataAndBacksidesWithClones = [
+    ...cardDataAndBacksides,
+    ...cardDataAndBacksides,
+  ];
+  const shuffledCardDataAndBacksidesWithClones = shuffle(
```
louisrli: shuffledCardDataBacksidesClones

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+  const cardBackSides = [];
+
+  return shuffledCardDataAndBacksidesWithClones.map((dataAndImage) => {
+    const { description } = dataAndImage[0];
+    const image = dataAndImage[1];
+    const isClone = cardBackSides.find(
+      (prevDescription) => prevDescription === description
+    )
+      ? true
```
louisrli: already something is wrong here when one thing returns a boolean but the other returns an empty string

louisrli: two different types

src/components/memoryCardBackSidesArr.jsx
```diff
@@ -0,0 +1,48 @@
...
+    ...cardDataAndBacksides,
+    ...cardDataAndBacksides,
+  ];
+  const shuffledCardDataAndBacksidesWithClones = shuffle(
+    cardDataAndBacksidesWithClones
+  );
+  const cardBackSides = [];
+
+  return shuffledCardDataAndBacksidesWithClones.map((dataAndImage) => {
+    const { description } = dataAndImage[0];
```
louisrli: so if you just make one file called cardspecs that has all the data of your cards, you can basically delete 80% of this code and just loop through one object. is there a particular reason card backs need to be separated from card fronts? in a real database, you would have them all stored together in a row/table (or at least join them together to get a single logical object)

src/components/memoryCardBackSidesArr.jsx
louisrli: i think you need to include in the PR what the game is, in addition to writing comments, again imagine you are your teammate or someone who joins this team trying to understand what's going on with the cloning thing

src/components/memoryCardBackSidesArr.jsx
tmansuroglu: To make flipping work, you need define 2 children(1 front ,1 backside) under `<ReactCardFlip></ReactCardFlip>`. Also, I didn't store them inside 1 object at first because im getting the data from i18n. Front image is same for all and backside images are repeated per language. But after getting data from i18n, I can create and object to hold everything.

louisrli: so there are a few options here that would be cleaner than what's currently here
* you store them inside 1 object but you stored the i18n translation keys instead of directly storing the data
* you create another virtual layer, processing after translation, to create objects that you can access (a class would be suitable here)

compare these two, and imagine you have to add more data eventually

```
    const { description } = dataAndImage[0];
    const image = dataAndImage[1];
    const anotherThing = dataAndImage[2];
```

vs code that looks like this (your goal)
```
console.log(cardData.image)
console.log(cardData.description)
console.log(cardData.title)
```

basically, if you can't define it upfront, you want to have a preprocessing step so that you (and your teammate, and your future self, who have to read / understand the code) can work with a cleanly defined cardData object

louisrli: i'm not understanding the cardflip thing and cloning. unless i'm misunderstanding the game. and even if you need to clone it, you shouldn't clone it at the time of preprocessing, you should clone it at the time the clone is needed (creating clones of data and passing them around gets extremely confusing).

```
const [isFlipped, setIsFlipped] = React.useState(false)

<div>
{!isFlipped && <CardFront data={cardData.front}/>}
{isFlipped && <CardBack data={cardData.back}/>}
</div>
```

louisrli: with cloning at usage time:
```
const [isFlipped, setIsFlipped] = React.useState(false)
const cardDataClone = // clone card data
<div>
{!isFlipped && <CardFront data={cardData.front}/>}
{isFlipped && <CardBack data={cardDataClone.back}/>}
</div>
```
or something like this, if for some reason the data needs to be cloned

tmansuroglu: Im cloning the cards because the purpose of game is matching cards with same backside image. Similar example is [this](https://www.brightfocus.org/alzheimers/memory-games/match-pictures). But instead of making square disappear after clicking on different sqaure, I made them disappear after certain amount of time. So you click on card it flips and shows what is behind and after some time it flicks back to show front side.

tmansuroglu: First I will create an object that holds everything. And I will separate object creating step to a different file to make it obvious.

tmansuroglu: Also clones actually are not identical. They have different keys. But I will try to implement what you suggested. I can implement it by lifting up `selectedCards`

src/components/memorygame/cardInitialStateGenerator.jsx
```diff
@@ -0,0 +1,65 @@
+import * as invariant from "invariant";
+import CARD_BACKSIDE_IMAGES from "./memoryCardBackSideImages";
+import shuffle from "./shuffle";
+
+/**
+ * cardDataArr example:
+ * [{description:"...",link:"...",phrase:".."},{description:"...",link:"...",phrase:".."},...]
```
louisrli, line 7: nicely done. if we were using typescript, we wouldn't need to do this because the types would be clearly declared, but in javascript we have to do things like this. this is also not ideal because these comments don't age well in case you change the structure but there's not much that can be done about that

src/components/memorygame/cardInitialStateGenerator.jsx
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * CARD_BACKSIDE_IMAGES example:[image1,image2,...]
+ *
+ * Game example: https://www.brightfocus.org/alzheimers/memory-games/match-pictures
+ * Instead of squares this game uses cards.
+ *
+ * This function prepares the statesArr that sets the structure of game board
+ */
+const cardInitialStateGenerator = (cardDataArr, NUM_REQUIRED_MATCHES) => {
```
louisrli: fn parameters should never be ALL_CAPS, numRequiredMatches (it doesn't know whether or not its taking a constant, it doesnt care, teh function just knows it accepts a number)

src/components/memorygame/cardInitialStateGenerator.jsx
```diff
@@ -0,0 +1,65 @@
...
+        ...card,
+        cardKey,
+        isFlipped: false,
+        isMatched: false,
+        img: { src: imgSrc, imgKey: imgName },
+      },
+    ];
+  });
+
+  // non destructive shuffle
```
louisrli: can delete this comment, it just needs to be in the declaration of shuffle (if i see this without the comment i assume its nondestructive because you return it)

src/components/memorygame/shuffle.jsx
```diff
@@ -0,0 +1,11 @@
+// Check Fisher–Yates shuffle algorithm https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
+const shuffle = (arr) => {
```
louisrli, line 4: nondestructive shuffle goes here; think about whether it makes more sense to comment "non-destructive shuffle" above every call to this function or in the place the function is declared (if i used htis 100x in my codebase and wrote it above every time, that would seem a bit tedious)

src/containers/MemoryGame/index.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import Navbar from "../../components/Navbar";
+import "./index.css";
+
+const MemoryGame = () => {
+  return (
+    <div className="memoryGameContainer">
+      <Navbar />
```
louisrli: i don't think you should be having to declare the Navbar on every page...seems wrong. it should be declared in a top-level component, and using routing the thing you want to show should show up.

src/components/memorygame/cardInitialStateGenerator.jsx
```diff
@@ -0,0 +1,65 @@
...
+    // clonedImgArr[index] is a directory
+    const imgSrc = clonedImgArr[index];
+    const imgNameParts = clonedImgArr[index].split("/");
+    const lastNamePart = imgNameParts[imgNameParts.length - 1];
+    // lastNamePart is something like acidRain.234234
+    const imgName = lastNamePart.split(".")[0];
+    // unique cardKey and same imgKey are needed for card match condition
+    // card keys of matching images should be like "acidRain0", "acidRain1"...
+    let cardKey = "";
+    const numOfSameCardKeys = cardStates.filter(
```
louisrli: ok...it took me a while to get this code, its doing something tricky which is that, as it goes through the forEach and builds more cards, then numOfSameCardKeys gets incremented, almost implicitly (if there are 6 acid rains and i'm through 3 of them, then this finds the previous 2).

i would call this variable numSameCardKeysSoFar and add a comment that this does the same thing as incrementing by doing this search

louisrli (approved):
generally good, minor comments pls fix before merging

src/components/memorygame/cardInitialStateGenerator.jsx
tmansuroglu: Added

src/containers/MemoryGame/index.jsx
tmansuroglu: fixed

src/components/memorygame/shuffle.jsx
tmansuroglu: fixed

src/components/memorygame/cardInitialStateGenerator.jsx
tmansuroglu: fixed

src/components/memorygame/cardInitialStateGenerator.jsx
tmansuroglu: fixed

tmansuroglu:
@majdajroudi @ssinikoglu @ibrahimaljarad  guys i need review for approval.

src/components/memoryCardBackSides.jsx
majdajroudi: This is an old comment that I wrote and never submitted, please ignore it

majdajroudi (approved):
very well documented. I will use your shuffle function in my PR

tmansuroglu:
@Ammar-64 can you merge?

PR #29, opened 2020-11-07
Memory game game card component

tmansuroglu:
#22 Please review . Test fails because i need to merge this with #25 . No screenshot because component is not meaningful alone.


src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState, createRef } from "react";
+import ReactCardFlip from "react-card-flip";
+import { Col, Card } from "antd";
+import FrontImage from "../../images/memoryGameFrontImage.png";
+import successSymbol from "../../images/memoryGameSuccessSymbol.svg";
+import "./index.css";
+
+const GameCards = ({ setMatchedCardIndexes, refArr, cardsArr }) => {
```
ssinikoglu: Instead of naming as GameCards, specifying game's name can be better.

tmansuroglu: Since it is inside `MemoryGameCards` component, i thought it is okay to call this `const` as `GameCards `. Because otherwise i have to use the word `memory` everywhere.

ssinikoglu: Thanks for your explanation Tarkan.

louisrli: i'd still name it MemoryGameCards. its better for readability, as sule said because you have a bunch of games, if you are scanning through files

louisrli: pretend you don't know this code for a second and read line 8. it is your first time to read this code. what are these props supposed to mean?

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const [selectedCards, setSelectedCards] = useState([]);
```
louisrli: pretend you don't know this code for a second and read this line. it is your first time to read this code. how are cards represented? are they strings?

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const initialStates = {};
+  cardsArr.forEach((card) => {
+    initialStates[card.key] = false;
+  });
+  const [cardStates, setCardStates] = useState(initialStates);
+
+  const handleClick = (curKey, index) => {
+    const flippedState = { [curKey]: true };
+    setCardStates((prevState) => ({ ...prevState, ...flippedState }));
+    setSelectedCards((prevState) => [...prevState, { key: curKey, index }]);
```
louisrli: from my experience coding, i will again advise that working with indexes in general will make you very, very miserable in terms of bugs and debugging.

it'd be better to have an imageKey and then a cardKey 
imageKey = "eelsInWater"
cardKey = "eelsInWater0", "eelsInWater1"

even the concept is easier to reason about for readers in general than thinking of numbers, abstractly each card has an "imageKey" (identifies the image answer that it belongs to) or it has a "cardKey" (identifies this card uniquely)

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const prevSelectedCardsIndex = selectedCards.length - 1;
+    const prevKey = selectedCards[prevSelectedCardsIndex]
+      ? selectedCards[prevSelectedCardsIndex].key
+      : "";
+    const prevCardState = cardStates[prevKey];
+    if (
+      prevKey &&
+      // prevCardState makes sure that only flipped cards match
+      prevCardState &&
+      (curKey === `Clone of ${prevKey}` || prevKey === `Clone of ${curKey}`)
```
louisrli: you should never, ever, ever, ever check magic strings like this (again not one of those things that i would consider an exception), **extra especially* if they have spaces and capitalization (even `${prevKey}clone` would be borderline, at that point you should have a function called `makeCloneDescription` so that its consistent in all places)

1. it is prone to typos. no one is checking if you typed "Clone of" vs "clone of", and that will break your code, program defensively here and assume writing code like this opens up to such mistakes
2. if you want to change the description, you have to change it in at least 3 places. where you cloned the code, and these two, and any other places you used it. google "magic strings"

this is the reason why i am suggesting the selected card states to be an array of "imageKeys" representing what the answer is.

```
selectedKeys: ["eel", "eel"]
```

```
if (all selected image keys are the same) {
   setMatchedCardKeys(["eel-clone0", "eel-clone1"])
} else {
   // do whatever was not successful here
}
```


as mentioned, you should also consider how you would write this code if i asked you to make it so you had to match 10 of the same picture instead of 2 of the same picture. the code should be written so that this would be a one line change

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+      }, [1750]);
+    }
+  };
+
+  // Defines how cards will be displayed on screen
+  return cardsArr.map((cardBackImage, index) => {
+    const imageKey = cardBackImage.key;
+    const flipState = cardStates[imageKey];
+    // refArr is created to manipulate style.display of success card and flip card
+    refArr.push(createRef());
```
louisrli: some terms to google, but generally when you use map() you should not have it affect anything outside the map. this is called side-effecting, it violates JS functional programming which is what map is intended to do. your code works, technically, but it kinda destroys the intended use of map, which is to yield a pure result.

instead, if you are going to side-effect like lin 53, do
```
const cardParts = [];
for (stuff in cardsArr) => {
   // Do this first to show that its a side effect happening
   refArr.push(createRef());

   cardParts.push(<JsxStuff>)
}
```

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+          />
+        </Card>
+
+        <ReactCardFlip isFlipped={flipState} flipDirection="vertical">
+          <Card
+            onClick={() => handleClick(imageKey, index)}
+            className="memoryCard"
+          >
+            <img
+              src={FrontImage}
```
louisrli: camelcase

src/components/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+              preview={false}
+              alt="green question mark"
```
louisrli: capitalization

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+// matchedCards is an array and used to display related facts on screen.
+// setCompleted and setRedirect are boolean
+const MemoryGameCards = ({
+  setMatchedCards,
+  matchedCards,
+  setCompleted,
+  setRedirect,
+}) => {
+  const [t] = useTranslation();
+  const NUM_REQUIRED_MATCHES = 2;
```
louisrli: move to line 9

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+  const cardsDataArr = [
+    ...t("memoryGame.cards", {
+      returnObjects: true,
+    }),
+  ];
+  // cardStates is an array of objects with multiple properties
+  const [cardStates, setCardStates] = useState(
+    cardInitialStateGenerator(cardsDataArr, NUM_REQUIRED_MATCHES)
```
louisrli, line 32: an initial state generator is a good pattern as well, as long as its clear theres a separate step from the "ground truth" card data

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+  );
+  // selectedCards is an array of clicked card states(object).
+  const [selectedCards, setSelectedCards] = useState([]);
+
+  const handleClick = (currentCardState) => {
+    const copyOfCardStates = [...cardStates];
+    const currentIndex = cardStates.indexOf(currentCardState);
```
louisrli: this seems quite dangerous to do .indexOf currentCardState when its an object. you're depending on it doing it by reference of the pointer to the object. i assume it works now but it could be the cause of some relaly tricky bugs, because generally you dont want to compare equality on objects

if there is a unique key that identifies the card state you want, i would write it more defensively using findIndex

cardStates.findIndex(state => state.key === currentCardState.key)

also currentCardState should probably be renamed clickedCardState, if i'm understanding this code correctly

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+    copyOfCardStates[currentIndex].isFlipped = true;
+    setCardStates(() => [...copyOfCardStates]);
+    setSelectedCards((prevState) => [...prevState, currentCardState]);
+  };
+
+  useEffect(() => {
```
louisrli: given this is quite a long useeffect function, it could use a comment above it. for example i looked at the dependencies but i don't quite understand when the effect is supposed to run (and this would be nice to know immediately before i read the code that is supposed to run)

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+    if (selectedCards.length === NUM_REQUIRED_MATCHES) {
```
louisrli: its nice to avoid long blocks of indentation when possible.

if (selectedCards.length !== NUM_REQUIRED_MATCHES) {
return
}

then the rest of the code can be intended two lines left

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+      const lastNCards = selectedCards.slice(-NUM_REQUIRED_MATCHES);
```
louisrli: i'm guessing, i've gotten this far, that this code checks if the user guessed the matches correctly. a comment to that effect above line 40 would be good

louisrli: isn't the length of selected cards REQUIRED_MATCHES? then the slice seems unnecessary. i'm confused why theres a variable called lastNCards to check for matches cuz wouldn't yo ualways just be checking the selectedcards for matches?

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+      const curCard = lastNCards[lastNCards.length - 1];
```
louisrli: the current card of what. what is it current for? isn't matching just checking if all of the cards in the sequence are the same?

// The last card of last N cards represents X

this is the comment that is needed

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+      const matchedCardArr = [
+        ...lastNCards.filter(
+          (prevCard) =>
+            prevCard.id !== curCard.id &&
```
louisrli: can you move these booleans into user readable variables. when you have long conditions a reader should be able to read and understand it in 1 second

for example you can move all 4 of these into one condition or function called, `doesMatchLastCard` or something like this. then you have .filter(prevCard => isSiblingOfCurrentCard(prevCard))

(but i don't know what ID means, why do they have to have the same ID?)

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+            prevCard.img.imgKey === curCard.img.imgKey &&
+            prevCard.isFlipped &&
+            curCard.isFlipped
+        ),
+        curCard,
+      ];
+
+      const copyOfCardStates = [...cardStates];
+      if (matchedCardArr.length === NUM_REQUIRED_MATCHES) {
```
louisrli: i will stop reading this code here because thsi whole useeffect thing needs more comments sprinkled throughout it for what's trying to be achieved in each section. think of comments as an outline, i should be able to read only the comments and know what is about to happen in each section

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+        });
+        setSelectedCards([]);
+        setTimeout(() => {
+          setCardStates(() => [...copyOfCardStates]);
+        }, 750);
+      }
+    }
+  }, [cardStates, selectedCards]);
+
+  useEffect(() => {
```
louisrli, line 112: comment above

// Checks for victory conditoin

louisrli: or something like this

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+    if (matchedCards.length === cardStates.length) {
+      setCompleted(true);
+      Modal.success({
+        content: "You have completed the game!",
+        onOk: () => setRedirect(true),
```
louisrli: if setRedirect is coming from the parent consider instead passing the child a function like `initiateRedirect` or something like this, since you are never going to do setRedirect(false), this component should have no knowledge of whether it requires a boolean or not. it just knows "hey i have the ability to cause a redirect"

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,134 @@
...
+      });
+    }
+  }, [matchedCards]);
+  return cardStates.map((cardState) => {
+    return (
+      <Col lg={4} key={cardState.cardKey}>
+        {cardState.isMatched ? (
```
louisrli: this ternary is too long, let the code read like `cardState.isMatched ? variable0 : variable1` and factor out these big JSX into variables between line 93 and 94

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

tmansuroglu: `lastNCards `was pointless and removed. Because when there is enough cards in selectedCards ,condition gets executed and in the end selected cards get emptied. PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `lastNCards ` removed and other comments added. PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `id` id is not being used anymore. function created . PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `setRedirect ` is removed. PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `true `condition of ternary is moved to a const. But false condition, is still there . It is `ReactCardFlip` component. When I put it inside a new component and store it to a const it doesn't update properly. I will let you know when I fix it.

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: More comments added PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+import questionMark from "../../../images/memoryGameFrontImage.png";
+import memoryGameSuccessSymbol from "../../../images/memoryGameSuccessSymbol.svg";
+import cardInitialStateGenerator from "../cardInitialStateGenerator";
+
+const NUM_REQUIRED_MATCHES = 2;
+// game example: https://www.brightfocus.org/alzheimers/memory-games/match-pictures
+// instead of squares this game uses cards.
+// onClick card flips to show what is behind the card.
+// matchedCards is an array states that holds successfully matched card states
+// setCompleted is boolean
```
louisrli: line 15/16 can be compacted into one line

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+import { Col, Card, Modal } from "antd";
+import { useTranslation } from "react-i18next";
+import questionMark from "../../../images/memoryGameFrontImage.png";
+import memoryGameSuccessSymbol from "../../../images/memoryGameSuccessSymbol.svg";
+import cardInitialStateGenerator from "../cardInitialStateGenerator";
+
+const NUM_REQUIRED_MATCHES = 2;
+// game example: https://www.brightfocus.org/alzheimers/memory-games/match-pictures
+// instead of squares this game uses cards.
+// onClick card flips to show what is behind the card.
```
louisrli: onClick sounds like a function that is a prop to this component. "Card flips on click to show what is behind."

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+  useEffect(() => {
+    // only triggered if "NUM_REQUIRED_MATCHES" amount of cards are clicked
+    if (selectedCards.length !== NUM_REQUIRED_MATCHES) {
+      return;
+    }
+    // match condition of selectedCards will be based on card that is added last to selectedCards
+    const lastCardState = selectedCards[selectedCards.length - 1];
+    const matchesOfLastCardState = selectedCards.filter((prevCard) =>
+      doesMatchLastCard(prevCard, lastCardState)
+    );
```
louisrli: rename doesMatchLastCard to:
`areCardsMatching`

with a comment
```
// Checks if two cards have a matching image.
```

you should also be able to delete line 55/56 the flipped conditions. this function doesn't care if its flipped; all selectedCards should be flipped by the time its in line 63 as far as i know. its not possible to be a selected card and not flipped, right? (duplicate checks like these lead to bugs)

```
// Check all cards correspond to the same image.
selectedCards.every((c) => {
   return areCardsMatching(c, selectedCards[0])
})
```

then you can clean up this code a bunch and delete lastCardState, matchesOfLastCardState, matchedCardsArr

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+        };
+      });
+      // to display 1 fact per match,
+      // only one of the matched cards(selectedCards[0]) should be passed as state
+      setMatchedCards((prevState) => [...prevState, selectedCards[0]]);
+      setCardStates(() => [...copyOfCardStates]);
+    }
+    // if there is no match or not enough matches, flipped cards get
+    // closed by setting isFlipped to false
+    else if (NUM_REQUIRED_MATCHES === selectedCards.length) {
```
louisrli: this condition is trivially true, this should just be an else-condition. you already return early if its not true

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+    if (selectedCards.length !== NUM_REQUIRED_MATCHES) {
+      return;
+    }
+    // match condition of selectedCards will be based on card that is added last to selectedCards
+    const lastCardState = selectedCards[selectedCards.length - 1];
+    const matchesOfLastCardState = selectedCards.filter((prevCard) =>
+      doesMatchLastCard(prevCard, lastCardState)
+    );
+    // matchesOfLastCardState doesnt include last card
+    const matchedCardsArr = [...matchesOfLastCardState, lastCardState];
```
louisrli: you don't need this variable at all, when matchedCardsArr.length == NUM_REQUIRED_MATCHES its just selected cards. the selected cards and the matched cards are the same

you should be able to do everything with selectedCards

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+      // to display 1 fact per match,
+      // only one of the matched cards(selectedCards[0]) should be passed as state
+      setMatchedCards((prevState) => [...prevState, selectedCards[0]]);
+      setCardStates(() => [...copyOfCardStates]);
+    }
+    // if there is no match or not enough matches, flipped cards get
+    // closed by setting isFlipped to false
+    else if (NUM_REQUIRED_MATCHES === selectedCards.length) {
+      selectedCards.forEach((card) => {
+        // current index is needed to keep the order of cards displayed
```
louisrli: this comment is wrong

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+          isMatched: true,
+        };
+      });
+      // to display 1 fact per match,
+      // only one of the matched cards(selectedCards[0]) should be passed as state
+      setMatchedCards((prevState) => [...prevState, selectedCards[0]]);
+      setCardStates(() => [...copyOfCardStates]);
+    }
+    // if there is no match or not enough matches, flipped cards get
+    // closed by setting isFlipped to false
```
louisrli: flipped cards reset isFlipped to false

change the second half of the comment

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+        copyOfCardStates[targetIndex] = {
+          ...copyOfCardStates[targetIndex],
+          isFlipped: false,
+        };
+      });
+      // this gives user 750ms to see failed match by delaying setting isFlipped to false.
+      // Otherwise game gets too punishing
+      setTimeout(() => {
+        setCardStates(() => [...copyOfCardStates]);
+      }, 750);
```
louisrli: magic number, move to top, give useful name

src/components/memorygame/MemoryGameCards/index.jsx
louisrli: at line 136

```
let cardBody;
if (cardState.isMatched) {
cardBody = successCard
} else {
cardBody = <ReactFlip>
}

{cardBody}
```

louisrli (commented):
pretty close just minor stuff

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: oh right. Before I had automatic unflip mechanism. That's why I still have flip condition. Fixed it.

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `selectedCards `holds clicked cards. `matchedCardsArr` holds only successfuly matched cards. The condition you mentioned checks if match is made. For example imagine that you need 3 matches per image(`const NUM_REQUIRED_MATCHES=3`) and game board is like ` a,a,a,b,b,b,c,c,c`. if you click on `c,a,a` in this order. `selectedCards` will hold `c,a,a` and `matchedCardsArr `will hold `a,a`. So if condition is `(selectedCards.length === NUM_REQUIRED_MATCHES)` , a wrong set of cards will have `isMatched:true`. Also, I can't use `selectedCards.length == NUM_REQUIRED_MATCHES` because there is `if (selectedCards.length !== NUM_REQUIRED_MATCHES) {
            return; }` above. Correct way of checking `if (selectedCards.length === NUM_REQUIRED_MATCHES)` would be `else`. But inside `else` i still have to check `matchedCardsArr.length ` .

tmansuroglu: I should clarify things better in comments. I'm having extremely difficult time making things clearer.

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: `.every` is a useful method. I implemented and adjusted the code accordingly PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

tmansuroglu:
I tried to improve readability by separating some code from stacked useEffect.

ibrahimaljarad (approved):

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,163 @@
...
+import questionMark from "../../../images/memoryGameFrontImage.png";
+import memoryGameSuccessSymbol from "../../../images/memoryGameSuccessSymbol.svg";
+import cardInitialStateGenerator from "../cardInitialStateGenerator";
+
+const NUM_REQUIRED_MATCHES = 2;
+// game example: https://www.brightfocus.org/alzheimers/memory-games/match-pictures
+// instead of squares this game uses cards.
+// Card flips on click to show what is behind.
+// matchedCards is an array states that holds successfully matched card states
+// setCompleted is true when all possible matches made
```
louisrli: set completed is a function, so the sentence that it is true (a boolean) doesn't make sense

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -0,0 +1,163 @@
...
+      targetCardState.isMatched = !targetCardState.isMatched;
+      setCardStates(() => [...copyOfCardStates]);
+      // matchedCards hold successfuly matched card states and uses them to display facts
+      // to prevent fact duplication only 1 of the matched cards should be passed into matchedCards
+      setMatchedCards((prevState) => [...prevState, selectedCards[0]]);
+      // isFlipped is false as default. When it is true backside of card become visible
+    } else if (property === "isFlipped") {
+      targetCardState.isFlipped = !targetCardState.isFlipped;
+      setCardStates(() => [...copyOfCardStates]);
+    }
```
louisrli: else throw an error.. defensively program here. otherwise if someone passes the wrong thing or makes a typo in property your code will simply run

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: You are right. Also, it should be `setIsCompleted`.

// `setIsCompleted` is a function that is being used to set `isCompleted` as boolean state when all possible matches made.

I think this is better

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: ` else {
            throw new Error('Invalid property was passed to changeCardStatePropertyToOpposite');
        }`

tmansuroglu: PTAL

src/components/memorygame/MemoryGameCards/index.jsx
tmansuroglu: PTAL

louisrli (approved):

tmansuroglu:
I will meet with Ammar to solve test problems. Then this PR will be mergable.

tmansuroglu:
@Ammar-64 ready to merge

src/components/memorygame/MemoryGameCards/index.jsx
```diff
@@ -78,9 +78,7 @@ const MemoryGameCards = ({ setMatchedCards, matchedCards, setIsCompleted }) => {
       targetCardState.isFlipped = !targetCardState.isFlipped;
       setCardStates(() => [...copyOfCardStates]);
     } else {
-      throw new Error(
-        "Invalid property was passed to changeCardStatePropertyToOpposite"
-      );
+      throw new Error("invalid property");
```
louisrli: why was this made to be less descriptive?

tmansuroglu: Oh I guess I reverted some changes by mistake. I will take a look

tmansuroglu: PTAL

tmansuroglu:
> as i recommended to majd too, you want to think in the future how you would implement the game logic perhaps with classes or more subcomponents
> 
> the ideal way to implement this would be probably be:
> 
> * use redux to hold the game state (google "game state" to understand htis concept if needed)
> * have separate components for each Card
> 
> all your game logic will then be in redux and you will basically have extremely clean components (your code works right now but there's pretty beastly logic in MemoryGameCards)
> 
> we generlaly want to separate UI logic from "business logic" (you can google this too if you're curious about this concept) so in the ideal world a component just knows "hey i was clicked and i'm gonna call redux flip()" without knowing much else about what happens.

@louisrli 
Totally makes sense. I wish I knew this before I started building the game. I think we should create an issue for this. So, redux and UI-logic separation can be implemented when everything else is finished. For now, I believe it is better to proceed with current code. 

I thought I used subcomponent appropriate way. Now, I realize that I could separate code a little bit more. But overall code is not that long. So, I don't think that readability is impaired heavily. But, I'm also having difficult time to look at things from a reviewers perspective when I'm coding. Either I suck at it or it is an experience thing. So, let me know if I'm delusional :D

tmansuroglu:
@louisrli Also, if everything is ok can you merge code?

PR #41, opened 2020-11-10
Image puzzle game - PuzzleGrid component

majdajroudi:
I fixed the problem of commits and created a new branch for this new PR. Please let me if I have to close the other PR.

This PR includes the general structure of the game along with most of the _CurrentImage_ component built. _CurrentImage_ component currently displays a group of divs ordered from 0 to 5 that display a text implying that the puzzle was solved when put in order.

_PuzzleGameContainer_ component will have a list of images where the user can choose one to solve as a puzzle.

### Game Description:
The game is a simple image puzzle game with multiple images to be solved, the tiles of the puzzle are ordered vertically where each row is one tile (as shown in the image attached). The user can choose which image they want to solve. The content of the page will be a card divided into two sections, upper part is list of images and bottom part is the current image the user is solving. The first image in the list is the one in the current image space by default, when the user solves the puzzle the image that was solved is displayed as a whole. 

### PR Summary:
This is PR consists of three components _PuzzlePage_, _PuzzleSpace_ and _CurrentImage_:
- **PuzzlePage:** this component creates the general layout of the page, where it will have the Navbar, Content and Footer

- ~~**PuzzleSpace:**~~ **PuzzlePageContent** this component can be considered as the content of the page, it includes a card (imported from antd) and the card is divided into two sections. The upper section consists of the list of images that user can choose to solve their puzzle, and the bottom section consists of the tiles of the image that is currently being solved. 
_**Update:**_ PuzzlePage and PuzzleSpace were removed from this PR due to reviewers' request and will be added using another PR

- ~~**CurrentImage**~~ **PuzzleGrid:** This component shows 6 tiles of the image, a user can swap tiles by clicking on one tile to select and click on the other to swap their places. A function called _checkSolution_ is called every time a couple of tiles is swapped which checks if the id of each tile is equal to its index in the array of tiles, if _true_ then the full image is displayed.
Screenshot from 2020-11-11 00-00-47


src/components/CurrentImage/index.css
```diff
@@ -0,0 +1,16 @@
+.puzzlePage__puzzleBox__CurrentImage {
```
louisrli: currentImage camelcase

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+import { Col, Row } from "antd";
+import "./index.css";
+
+const CurrentImage = () => {
+  const [generatedTiles, setGeneratedTiles] = useState([]);
+  const [tileSelected, setTileSelected] = useState({
+    id: null,
+    selected: false,
+  });
+  const [isSolved, setIsSolved] = useState(false); // if unsolved it displays the tiles and when solved it displays the full image
```
louisrli: comments go above the lıne

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const generateTiles = () => {
+    let newTilesArray = [];
+    for (let i = 0; i < 6; i++) {
```
louisrli: magic number

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    for (let i = 0; i < length; i++) {
+      let random = Math.floor(Math.random() * (length - 1));
+      let randomItemArray = newTilesArray.splice(random, 1);
+      newTilesArray.push(randomItemArray[0]);
+    }
+    setGeneratedTiles(newTilesArray);
+  };
+
+  useEffect(() => {
+    generateTiles();
```
louisrli: you have a variable called generateTiles and generatedTiles, rename them somehow so its less ambiguous to readers and less prone to cause mistakes

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    newTiles[index1] = tile2;
+    newTiles[index2] = tile1;
+    newTiles.map((tile) => (tile.selected = false));
+    setGeneratedTiles(newTiles);
+    setTileSelected({ id: null, selected: false });
+  };
+
+  const checkSolution = () => {
+    const isCorrect = generatedTiles.reduce((curr, tile, i) => {
+      if (tile.id !== i) {
```
louisrli: this reduce is fairly unreadable. on top of that the order of a reducers arguments are "accumulator, current element in reduce, index", but the first argument in the impl is called "curr" (which is the accumulator)

consider rewriting it, adding a comment explaining how the isCorrect computation works, using a normal for-loop, or maybe all of the above

also in this impl i think line 65/66 and line 68/69 can just be collapsed into one line, return true or return false

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+      generatedTiles[tileIndex] = {
+        ...generatedTiles[tileIndex],
+        selected: true,
+      };
+    }
+  };
+
+  const swapTiles = (id1, id2) => {
+    let newTiles = [...generatedTiles];
+    let index1 = generatedTiles.findIndex((tile) => tile.id === id1);
```
louisrli: variables should be 0 indexed. id0, id1 fix everywhere

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+    let index2 = generatedTiles.findIndex((tile) => tile.id === id2);
+    let tile1 = { ...generatedTiles[index1] };
+    let tile2 = { ...generatedTiles[index2] };
+    newTiles[index1] = tile2;
+    newTiles[index2] = tile1;
+    newTiles.map((tile) => (tile.selected = false));
+    setGeneratedTiles(newTiles);
+    setTileSelected({ id: null, selected: false });
```
louisrli: this needs some documentation. maybe its called setSelectedTile? or setTileSelected? and what does it mean for the object to have "selected: true/false". the variable is called setTileSelected which implies the tile is selected already. is this a special single tile (i assume yes because it holds one object). but can't the user select two tiles? (i shouldn't have to read all the code to understand this, it should be explained at the time you declare the state by a comment)

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const [isSolved, setIsSolved] = useState(false); // if unsolved it displays the tiles and when solved it displays the full image
+  const generateTiles = () => {
+    let newTilesArray = [];
+    for (let i = 0; i < 6; i++) {
+      let newTile = { id: i, selected: false }; // These tiles are puzzle pieces
+      newTilesArray.push(newTile);
+    }
+    let length = newTilesArray.length;
+    for (let i = 0; i < length; i++) {
+      let random = Math.floor(Math.random() * (length - 1));
```
louisrli: needs comment above this code block abt what's going on. also you can inline the variable on line 18

src/containers/PuzzlePage/index.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import PuzzleSpace from "../../components/PuzzleSpace";
+import Navbar from "../../components/Navbar/index";
+
+const PuzzlePage = () => {
```
louisrli: there are 4 components in this PR. i don't remember what happened in the last PR but still when there are 4 components in one PR it seems to me that it can easily be split into smaller parts. i didn't review anything after the first component. could you make one PR per component or at most two small components in the same PR please

src/components/CurrentImage/index.css
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: There was a bug in this function that I realized after your comment, I fixed the bug and added the comments PTAL

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: I added some explanation regarding this state at the declaration and a more detailed explanation at the implementation in handleSelect function PTAL

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/containers/PuzzlePage/index.jsx
majdajroudi: Since I only have 2 big components I removed one of the big ones for which I will make a new PR. but I kept the PuzzleSpace and PuzzlePage since they are relatively very small components and of course the main one currentImage component. I hope that's ok, I kept the small ones because I want to see in the browser whether what i'm doing is correct or not

src/components/CurrentImage/index.jsx
louisrli: fix, move to a constant

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
+import React, { useState, useEffect } from "react";
+import { Col, Row } from "antd";
+import "./index.css";
+
+const CurrentImage = () => {
```
louisrli: i'm not really understanding why this component is called currentimage

louisrli: are you saying because its a puzzle that forms a whole image? i think there's probably better names for this if that's the case. i would ask you to consider what comes to mind when you tell someone the phrase 'current image'

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const [generatedTiles, setGeneratedTiles] = useState([]);
+  // tileSelected state shows if there is a tile selected and which one is selected if any
```
louisrli: if this comment is true, can't tileSelected simply be of a type string | null?

if tileSelected is null, no tile is selected
if a tile is selected, the id is the id of the tile

unless it's possible for id to be true but selected to be false?

src/components/CurrentImage/index.jsx
louisrli: google magic number and see how most people do this, or ask tarkan

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    selected: false,
+  });
+  // if unsolved it displays the tiles and when solved it displays the full image
+  const [isSolved, setIsSolved] = useState(false);
+  const tilesGenerator = () => {
+    let newTilesArray = [];
+    // for loop to generate tiles where tiles are puzzle pieces
+    // the number 6 is chosen randomly
+    for (let i = 0; i < 6; i++) {
+      let newTile = { id: i, selected: false };
```
louisrli: is it possible for multiple tiles to be selected at once? if not, you probably don't want each tile to have its own selected property. but the answer to this question should be clarified above the usage of the useState for this too.

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const swapTiles = (id0, id1) => {
+    let newTiles = [...generatedTiles];
+    let index0 = generatedTiles.findIndex((tile) => tile.id === id0);
+    let index1 = generatedTiles.findIndex((tile) => tile.id === id1);
+    let tile0 = { ...generatedTiles[index0] };
+    let tile1 = { ...generatedTiles[index1] };
+    newTiles[index0] = tile1;
+    newTiles[index1] = tile0;
+    newTiles.map((tile) => (tile.selected = false));
```
louisrli: you are doing a side-effect in a map, you should never do this again (google "what is functional programming", "what is side effects", etc.) probably what you want here is a foreach

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+  useEffect(() => {
+    tilesGenerator();
+  }, []);
+
+  useEffect(() => {
+    checkSolution();
+  }, [generatedTiles]);
+
+  const handleSelect = (currId) => {
+    // if there is a tile selected then clicked tile with the selected one
```
louisrli: i don't understand this sentence

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    // else set tileSelected.select to true and show the id of the tile
+    if (tileSelected.selected) {
+      const prevId = tileSelected.id;
+      swapTiles(prevId, currId);
+    } else {
+      setTileSelected({ id: currId, selected: true });
+      const tileIndex = generatedTiles.findIndex((tile) => tile.id === currId);
+      generatedTiles[tileIndex] = {
+        ...generatedTiles[tileIndex],
+        selected: true,
```
louisrli: so does selected never become false again after becoming true?

i think this whole file needs more comments at the top explaining what it's doing

my recommendation, write a comment at the top of the file that explains exactly what this game is, what clicking does, and what it should do and/or include an example

ask your teammate to read it and see if they understand what the purpose of the code is without reading any of the code (just this comment at the top of file)

if they understand, then request rereview

majdajroudi:
> This PR needs to be split into multiple parts. I will review when it is split

I removed _PuzzleSpace_ and _PuzzlePage_ components and will create another PR for them 
@tmansuroglu PTAL

src/components/CurrentImage/index.jsx
majdajroudi: You are right, I fixed the value of tileSelected to null | number, where the number represents the ID of the selected tile. I also changed the name of the state to selectedTile, which probably makes more sense. PTAL

src/components/CurrentImage/index.jsx
majdajroudi: no it's not possible, the value of selected is used for styling purposes, where the selected value should be highlighted. I added comments in that part of the code PTAL

src/components/CurrentImage/index.jsx
majdajroudi: I fixed it and added a better description PTAL

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+import "./index.css";
+
+// CurrentImage is the component that generates and includes the tiles of the puzzle
+// it generates 6 tiles ordered vertically on top of each other
+// each tile's background is 1/6 of the image. After the tiles are generated they are put
+// in an array and ordered randomly
+// a user can click on one tile to select and then click on another tile to swap their places
+// the user will swap the places of each tile to put it in its correct position
+// when all tiles are placed in their correct position the full image is displayed
+const CurrentImage = () => {
```
louisrli: this component should be called PuzzleGrid or PuzzleTileGrid or something like this

CurrentImage still seems to me like its the "final image" that they wanna do, not the actual thing that they are editing

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const [generatedTiles, setGeneratedTiles] = useState([]);
+  // selectedTile state shows if there is a tile selected and which one is selected if any
+  const [selectedTile, setSelectedTile] = useState(null);
+  // if unsolved it displays the tiles and when solved it displays the full image
+  const [isSolved, setIsSolved] = useState(false);
+  const tilesGenerator = () => {
```
louisrli: newline above here, its like a new paragraph

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    const newTilesArray = [];
+    // for loop to generate tiles where tiles are puzzle pieces
+    const numOfTiles = 6;
```
louisrli: magic number still

const NUM_OF_TILES = 6

and outside of the component under the imports

src/components/CurrentImage/index.jsx
majdajroudi: It does become false in the swapTiles function, I added more comments to the code and 2 teammates approved the games description at the top of the code PTAL

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+                  ? "selectedTile currentImage__tile"
+                  : "currentImage__tile"
+              }
+              onClick={() => handleSelect(tile.id)}
+            >
+              <img
+                className="currentImage__puzzledImage"
+                src={puzzledImage}
+                alt="tile from the image"
+                style={{ top: `-${tile.top}%` }}
```
louisrli: this is really weird...shouldn't each img be on a new row and simply stack on top of each other? i don't think this is necessary. but it should definitely be taken out, usually you don't pass CSS information around like this, so for me its a red flag thing

i would play with the CSS here.
```
<Row><img></Row>
<Row><img></Row>
<Row><img></Row>
<Row><img></Row>
<Row><img></Row>
```

by changing the css of Row, or using display:block on img, or surrounding the img in divs, you might be able to fix this. or add margins on rows.

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      let tileTopStyling = tileId * 100;
+      // value of selected is used for styling purposes where the selected tile should be highlighted
+      // value of top is used for fitting the image inside the tile
+      let newTile = { id: tileId, selected: false, top: tileTopStyling };
+      newTilesArray.push(newTile);
+    }
+    // for loop that generates a random and extracts the element with the index
+    // equal to the random number, then pushes that element to the end of newTilesArray.
+    for (let i = 0; i < newTilesArray.length; i++) {
+      let random = Math.floor(Math.random() * (newTilesArray.length - 1));
```
louisrli: if i'm understanding correctly, this is basically shuffling the tiles right? if so use a shuffle function or re-use tarkans

this code is pretty tricky to read and does something that is highly unrecommended, which is changing the length of a list while looping through it (you splice and then push back to it). still, it makes it hard to understand whats going on 

https://stackoverflow.com/questions/1637807/modifying-list-while-iterating

> I've been bitten before by (someone else's) "clever" code that tries to modify a list while iterating over it. I resolved that I would never do it under any circumstance.

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      let randomItemArray = newTilesArray.splice(random, 1);
+      newTilesArray.push(randomItemArray[0]);
+    }
+    // Set the randomized array as generated tiles
+    setGeneratedTiles(newTilesArray);
```
louisrli: newTilesArray should probably just be called randomizedTilesArray

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const swapTiles = (id0, id1) => {
+    let newTiles = [...generatedTiles];
+    let index0 = generatedTiles.findIndex((tile) => tile.id === id0);
+    let index1 = generatedTiles.findIndex((tile) => tile.id === id1);
+    let tile0 = { ...generatedTiles[index0] };
+    let tile1 = { ...generatedTiles[index1] };
+    newTiles[index0] = tile1;
+    newTiles[index1] = tile0;
+    newTiles.forEach((tile) => (tile.selected = false));
+    setGeneratedTiles(newTiles);
```
louisrli: this state generatedtiles should be called "currentPuzzleState" or "currentPuzzleX" i believe. it represents the current state of the game board right that the user has modified right?

"generatedTiles" sounds like something computer generated. name is not clear that its related to what the user has modified.

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const checkSolution = () => {
+    // tiles are saved in an array indexed from 0 to n (6) and displayed in that order
+    // first condition checks if tile's id is equal to its index in the array and returns false/true
+    // since initial acc value is true, second condition checks if prev value is true and keeps its value
+    const isCorrect = generatedTiles.reduce((acc, curr, index) => {
+      if (curr.id !== index) {
+        return false;
+      } else if (acc === true) {
+        return acc;
+      }
```
louisrli: still seems like 85 to 89 can be rewritten 

its generally dangerous to have code that has a case that has no return value. here, you have an if and an else-if; two conditions. my question is, is there a third case where neither of those is true, and this code is just returning nothing? (the conditionals should be structured so i dont have to ask this question after i read it)

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    }, true);
+    setIsSolved(isCorrect);
+  };
+
+  return (
+    <Col className="currentImage">
+      {!isSolved ? (
```
louisrli: avoid beastly ternarry operators like this

const gameBody;

if (isSolved) {
 gameBody = 
} else {
gameBody = blah.map
}

louisrli:
> PuzzlePage: this component creates the general layout of the page, where it will have the Navbar, Content and Footer
> 
> PuzzleSpace: this component can be considered as the content of the page, it includes a card (imported from antd) and the card is divided into two sections. The upper section consists of the list of images that user can choose to solve their puzzle, and the bottom section consists of the tiles of the image that is currently being solved.
> Update: PuzzlePage and PuzzleSpace were removed from this PR due to reviewers' request and will be added using another PR
> 
> CurrentImage: This component shows 6 tiles of the image, a user can swap tiles by clicking on one tile to select and click on the other to swap their places. A function called checkSolution is called every time a couple of tiles is swapped which checks if the id of each tile is equal to its index in the array of tiles, if true then the full image is displayed.
> 

PuzzlePage: name is fine

PuzzlaSpace: space not a great word to describe things, since all things take up space. I'd call this PuzzleGameContainer, we usually use this to describe the supporting part of a page

CurrentImage: see my other comment here

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    checkSolution();
+  }, [generatedTiles]);
+
+  const handleSelect = (currId) => {
+    // if selectedTile's value is not equal to null (meaning there a tile seleted already)
+    // then when another tile is clicked swap both tiles' places
+    // else set the selectedTile to the id of the clicked tile
+    if (selectedTile !== null) {
+      const prevId = selectedTile;
+      swapTiles(prevId, currId);
```
louisrli: this code is looking much much better and much more readable

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+      setSelectedTile(currId);
+      const tileIndex = generatedTiles.findIndex((tile) => tile.id === currId);
```
louisrli: tileIndex -> clickedTileIndex

src/components/CurrentImage/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+  useEffect(() => {
+    tilesGenerator();
+  }, []);
+
+  useEffect(() => {
+    checkSolution();
+  }, [generatedTiles]);
+
+  const handleSelect = (currId) => {
```
louisrli: currId -> clickedTileId

louisrli (commented):
the code is generally looking good now

minor fixes

as a note, but not something that i'll make you do because i'm not that cruel, you would ideally be using OOP here. generally what people would do is define a class for the gameboard and define a tiles class.

then your code reads very nicely like this, for example:
```
gameBoard.swapTiles(tile0.id, tile1.id)
gameBoard.isSolved()
```

and basically it would clean up a lot of state/things in the core component. because components are meant to be UI logic usually (how do things go on the screen); so the cleanest separation comes from partitioning  the game logic from the UI logic using classes. game logic = how do we solve this puzzle; UI logic = lets render this component on the screen

src/components/CurrentImage/index.jsx
majdajroudi: The way i'm doing this is i'm placing the image inside a div and making the image 6 times bigger than the div then displaying the part that i want to display of that image. I think what you are saying is I should create a new div inside the `<Row>` and then add the CSS to that div instead of add it to the img. please correct me if i misunderstood

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: I know you said I should rename it, but instead I thought it's better to create a new array with the same values of newTilesArray but randomized, using the shuffle function Tarkan created. PTAL

src/components/CurrentImage/index.jsx
majdajroudi: PTAL

src/components/CurrentImage/index.jsx
majdajroudi: I fixed the issue and since there are only 2 way that the acc can be false, I included these two conditions at the top and for all the other conditions (else) the result would be true PTAL

src/components/CurrentImage/index.jsx
majdajroudi: I couldn't find the term "beastly ternary operator" but what I understood is that I shouldn't use the condition `!isSolved?` and should use `isSolved?` instead. PTAL and correct me if I misunderstood

src/components/PuzzleGrid/index.test.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import CurrentImage from "./index.jsx";
```
louisrli: currentimage -> puzzlegrid

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
+import React, { useState, useEffect } from "react";
+import { Col, Row } from "antd";
+import puzzledImage from "./../../images/habitat.jpg";
+import shuffle from "./../memorygame/shuffle";
+import "./index.css";
+
+const NUM_OF_TILES = 6;
+// PuzzleGrid is the component that generates and includes the tiles of the puzzle
```
louisrli, line 9: newline

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  const [isSolved, setIsSolved] = useState(false);
+
+  const tilesGenerator = () => {
+    const newTilesArray = [];
+    // for loop to generate tiles where tiles are puzzle pieces
+    for (let tileId = 0; tileId < NUM_OF_TILES; tileId++) {
+      let tileTopStyling = tileId * 100;
+      // value of selected is used for styling purposes where the selected tile should be highlighted
+      // value of top is used for fitting the image inside the tile
+      let newTile = { id: tileId, selected: false, top: tileTopStyling };
```
louisrli, line 36: > The way i'm doing this is i'm placing the image inside a div and making the image 6 times bigger than the div then displaying the part that i want to display of that image. I think what you are saying is I should create a new div inside the <Row> and then add the CSS to that div instead of add it to the img. please correct me if i misunderstood


yeah what i was saying is that you should try to use the normal flow of block ordering in css. to put it another way, when i see the "negative top" used to order 6 things on top of each other i immediately go like "that's a really strange way to achieve this." 

you don't need to use <Row> but if what i'm reading is correct, you want the image to be 6 times bigger than the parent div, i think you might want to do something like display: block on the img and make the parent div bigger. usually we want things to "make sense", so the child of a DOM element ebing 6 times bigger than its parent can be pretty weird

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+      setSelectedTile(clickedTileId);
+      const clickedTileIndex = currentPuzzleState.findIndex(
+        (tile) => tile.id === clickedTileId
+      );
+      // set the select value of the tile to true
+      currentPuzzleState[clickedTileIndex] = {
+        ...currentPuzzleState[clickedTileIndex],
+        selected: true,
+      };
+    }
```
louisrli, line 71: i think you need to do a setCurrentPuzzleState here. it's mutating the state in place which we shouldn't be doing (it won't cause a rerender).

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+      } else {
+        return acc;
+      }
+    }, true);
+    setIsSolved(isCorrect);
+  };
+
+  return (
+    <Col className="puzzleGrid">
+      {isSolved ? (
```
louisrli: beastly ternary operator = really big ternrary operator

like
```
X ? ASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSA : ASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSAASDFASDFADSFSA
```

it gets hard to read

```
let gridBody;
if (isSolved) {
  gridBody = <img>
} else {
  gridBody = // other thing
}

<Col>{gridBody}</Col>
```

louisrli (commented):
generally lgtm minus this top thing

louisrli:
I think I might be misunderstanding the top thing. is that what you're using to cut the images into 6 portions? how do you take current image and make it a puzzle?

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
+import React, { useState, useEffect } from "react";
+import { Col, Row } from "antd";
+import puzzledImage from "./../../images/habitat.jpg";
+import shuffle from "./../memorygame/shuffle";
```
tmansuroglu, line 4: I think it is better to move this under `src/components`

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+const NUM_OF_TILES = 6;
+// PuzzleGrid is the component that generates and includes the tiles of the puzzle
+// it generates 6 tiles ordered vertically on top of each other
+// each tile's background is 1/6 of the image. After the tiles are generated they are put
+// in an array and ordered randomly
+// a user can click on one tile to select and then click on another tile to swap their places
+// the user will swap the places of each tile to put it in its correct position
+// when all tiles are placed in their correct position the full image is displayed
+const PuzzleGrid = () => {
+  const [currentPuzzleState, setCurrentPuzzleState] = useState([]);
```
tmansuroglu, line 18: This needs explanation.

src/components/PuzzleGrid/index.jsx
tmansuroglu: `tileTopStyling ` in each iterations are 0,100,200,300,400,500 and what are they? top margin? height of top ? You should comment it

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+    for (let tileId = 0; tileId < NUM_OF_TILES; tileId++) {
+      let tileTopStyling = tileId * 100;
+      // value of selected is used for styling purposes where the selected tile should be highlighted
+      // value of top is used for fitting the image inside the tile
+      let newTile = { id: tileId, selected: false, top: tileTopStyling };
+      newTilesArray.push(newTile);
+    }
+
+    const randomizedTilesArray = shuffle(newTilesArray);
+    // Set the randomized array as generated tiles
```
tmansuroglu: unnecessary comment

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  useEffect(() => {
+    tilesGenerator();
+  }, []);
+
+  useEffect(() => {
+    checkSolution();
+  }, [currentPuzzleState]);
+
+  const handleSelect = (clickedTileId) => {
```
tmansuroglu: i think it should be called `handleClick` because select reminds form element

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+    // if selectedTile's value is not equal to null (meaning there a tile seleted already)
+    // then when another tile is clicked swap both tiles' places
+    // else set the selectedTile to the id of the clicked tile
+    if (selectedTile !== null) {
+      const prevId = selectedTile;
```
tmansuroglu: How can a tile be id

src/components/PuzzleGrid/index.jsx
tmansuroglu: I agree

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+      );
+      // set the select value of the tile to true
+      currentPuzzleState[clickedTileIndex] = {
+        ...currentPuzzleState[clickedTileIndex],
+        selected: true,
+      };
+    }
+  };
+
+  const swapTiles = (id0, id1) => {
```
tmansuroglu: `id0 `and `id1 `are not descriptive enough. same apply to similar namings.

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+    setCurrentPuzzleState(newTiles);
+    setSelectedTile(null);
+  };
+
+  const checkSolution = () => {
+    // tiles are saved in an array indexed from 0 to n (6) and displayed in that order
+    // first condition checks if tile's id is equal to its index in the array and returns false/true
+    // since initial acc value is true, second condition checks if prev value is true and keeps its value
+    const isCorrect = currentPuzzleState.reduce((acc, curr, index) => {
+      if (curr.id !== index || acc === false) {
```
tmansuroglu: `curr.id` and `index `are not descriptive. Are you comparing current position and correct position? If yes you should name them accordingly.

src/components/PuzzleGrid/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+                  ? "selectedTile puzzleGrid__tile"
+                  : "puzzleGrid__tile"
+              }
+              onClick={() => handleSelect(tile.id)}
+            >
+              <img
+                className="puzzleGrid__puzzledImage"
+                src={puzzledImage}
+                alt="tile from the image"
+                style={{ top: `-${tile.top}%` }}
```
tmansuroglu: Usually inline styling is not allowed. But this is short and variable dependent. @louisrli Is there a better option?

tmansuroglu (changes requested):
I'm sorry for late review.

majdajroudi:
> I think I might be misunderstanding the top thing. is that what you're using to cut the images into 6 portions? how do you take current image and make it a puzzle?

Yes, what i'm trying to achieve here is a cropping effect of the image, so I set overflow of the wrapping div to hidden, then make the image inside that div 6 times larger than the div, but since I want to show a specific portion of the image I move the image to the top in each tile by a percentage of the wrapping div's height (this percentage depends on the id of the tile). It's a trick that  I found on stackoverflow and I guess it might work with `bottom: (some positive value)` but I haven't tried that

src/components/PuzzleGrid/index.jsx
majdajroudi: You are right, but I kept like this for now since Sule is also using this function and it might cause a problem when our PRs are merged, I thought it's better to move it after all of our PRs are merged

src/components/PuzzleGrid/index.jsx
majdajroudi: I don't understand this comment

src/components/PuzzleGrid/index.jsx
majdajroudi: I want to keep the naming in this fn general, do you think a comment that explains what the fn does would make the whole function more descriptive? I have no problem with renaming but it's just a suggestion

src/components/PuzzleGrid/index.jsx
majdajroudi: Exactly, each tile will have it's own `top` value so I did it here. please let me know if there is an alternative

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

louisrli:
ok, in that case, the top thing is fine. but the comments need to be improved. as far as i can tell the word "crop" appears 0 times when describing the top thing

src/components/PuzzleGrid/index.jsx
tmansuroglu: `const prevId = selectedTile;`    how can a tile be equal to id

src/components/PuzzleGrid/index.jsx
tmansuroglu: i think renaming is better because id0 and id1 will not mean anything to the person who will fix/work on this stuff later.

src/components/PuzzleGrid/index.jsx
louisrli: you might be able to use n-th child but i'd probably prefer this inline styling here since its dynamically calculated

src/components/PuzzleGrid/index.jsx
louisrli: i would call these tileId0 and tileId1

one of the exceptions for being able to use variables that end with 0 and 1 which is basically when you have a function that takes two of the same type of thing but you don't really care what they are, just that they are a 0th and 1st instance of some thing

other examples might be comparator functions for sort `sort(t0, t1)` or `areDaysDifferent(day0, day1)` things like this

src/components/PuzzleGrid/index.jsx
louisrli: you have a variable called prevId and you assign it to selectedTile. the word tile implies that its an object. an id implies its a string

this is a good one to be careful of too, what you mean is selectedTileId

src/components/PuzzleGrid/index.jsx
louisrli: +1 someone please file an issue for the cleanup and make sure its done by end of sprint

louisrli: +1 to majd i mean, i would just use memorygame for now and move it after. or move it in a separate PR right now ( a really small one) and force everyone to merge

tmansuroglu: I created issue for it

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: PTAL

src/components/PuzzleGrid/index.jsx
majdajroudi: I added a detailed comment on the top in line 30 PTAL and let me know what you think

louisrli (approved):

louisrli:
lgtm

tmansuroglu (approved):
Nice work

majdajroudi:
PR approved PTAL @Ammar-64

Ammar-64:
> PR approved PTAL @Ammar-64 

@majdajroudi can you update with dev please

Issue #35, opened 2020-11-09
Contact page


ssinikoglu:
Build contact page according to figma design, create contact form, check netfliy form submission and connect it to firebase(optional), build tests, make it responsive.

PR #48, opened 2020-11-12
Contact us header component #35

ssinikoglu:
Please review. This PR consists of header component of Contact page. I will also make other PR's for ContactUsForm, NetlifyForm, also components for bottom part of the page and container.


majdajroudi (approved):
short and clean code, LGTM

src/components/ContactUsHeaderSection/index.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import "./index.css";
+import { useTranslation } from "react-i18next";
+
+const ContactUsHeaderSection = () => {
+  const { t } = useTranslation();
+  return (
+    <>
```
ibrahimaljarad: do you think you could wrap the element with a div? just a suggestion

src/locales/en.json
```diff
@@ -73,5 +73,11 @@
         }
       ]
     }
+  },
+  "contact": {
+    "headerSection": {
```
ibrahimaljarad, line 144: we have headersection key in the home page also, i know they are not related but i think its better to change the naming here

src/components/ContactUsHeaderSection/index.jsx
ssinikoglu: Thanks for your review Ibrahim. After your comment I tried adding a wrapper div and made some styling as well. PTAL

src/locales/en.json
ssinikoglu: Thanks for reviewing Ibrahim, I renamed the key according to your advice. PTAL

ibrahimaljarad (approved):

src/locales/ar.json
```diff
@@ -73,5 +73,11 @@
         }
       ]
     }
+  },
+  "contact": {
+    "contactUsHeaderSection": {
```
louisrli: I'd prefer if you can get this translation before you submit. normally we don't submit code that's half complete because there's a very high possibility someone will forget to do the other half. so let's get this in then I'll approve, rest of code is fine though

louisrli: contact.headerSection is fine here because the parent part already says contact

ssinikoglu: Thanks for reviewing! I have added Arabic translation. PTAL

ssinikoglu: Thanks for your review, I have changed the key name to "headerSection". PTAL

louisrli (approved):

PR #49, opened 2020-11-12
Contact us form and NetlifyForm components #35

ssinikoglu:
Please review. This PR consists of ContactUsForm and NetlifyForm components of Contact page. I have added NetlifyForm to the PR, because it is closely related with ContactUsForm and also is not meaningful on its own. I have made PR for header component of the page, and  I will also make other PR's for components for bottom part of the page and container.


src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+      contactFormValues.email.length > 0
+    ) {
+      setProcessing(true);
+      // Wait 2 seconds to simulate async delay
+      await new Promise((resolve) => setTimeout(resolve, 2000));
+      return true;
+    } else {
+      return false;
+    }
+  };
```
tmansuroglu: comment in line 15 should be moved to line 16

tmansuroglu: is `processing `used to create spinner ? You should clarify what it does

tmansuroglu: Does `submissionMessage` display something on screen when form is sent? You should clarify

tmansuroglu: in line 19 and 20, instead of using a magic number you should use a const. `MIN_USER_INPUT_LENGTH` maybe? I'm not good at naming things.

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+        formName="contact"
+        contactFormValues={contactFormValues}
+        beforeSubmit={beforeSubmit}
+        afterSubmit={afterSubmit}
+        automaticHoneypot={true}
+      >
+        <span className="contactUsForm__result">{submissionMessage}</span>
+        <Col>
+          <Col>
+            <label for="name" className="contactUsForm__content__label">
```
tmansuroglu: className="contactUsForm__contentLabel"

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+              {t("contact.contactUsForm.name")}
+            </label>
+          </Col>
+          <Col>
+            <input
+              type="text"
+              name="name"
+              value={contactFormValues.name}
+              onChange={handleChange}
+              className="contactUsForm__content__input"
```
tmansuroglu: className="contactUsForm__contentInput"

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+            />
+          </Col>
+        </Col>
+        <Col>
+          <Col>
+            <label for="email" className="contactUsForm__content__label">
```
tmansuroglu: className="contactUsForm__contentLabel"

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+              {t("contact.contactUsForm.email")}
+            </label>
+          </Col>
+          <Col>
+            <input
+              type="email"
+              name="email"
+              value={contactFormValues.email}
+              onChange={handleChange}
+              className="contactUsForm__content__input"
```
tmansuroglu: className="contactUsForm__contentInput"

tmansuroglu: You should check all class names

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+ * The base code that is used in both netlifyForm and index are implemented from https://github.com/jon-fm/react-ssg-netlify-forms-demo
+ */
+import React, { useState, useEffect } from "react";
+
+const NetlifyForm = ({
+  formName,
+  beforeSubmit,
+  afterSubmit,
+  contactFormValues,
+  children,
```
tmansuroglu: You should clarify use and types of these parameters inside comments

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+}) => {
+  // Keep things well-hidden during runtime (and from bots. They are exposed for netlify build bots.
```
tmansuroglu: Keep what hidden?

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  const [inNetlifyBuild, setInNetlifyBuild] = useState(true);
+  useEffect(() => {
+    setInNetlifyBuild(false);
+  }, []);
+
+  // Honeypot Input state for "Honeypot" field, which used Honeypot form field.
+  const [honey, setHoney] = useState("");
```
tmansuroglu: What is honeypot field. Is it certain type of form? You should put a link here as comment.

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  // For application/x-www-form-urlencoded, spaces are to be replaced by +, encodeURIComponent() replacement with an additional replacement of %20 with +. */
```
tmansuroglu: `encodeURIComponent() replacement with an additional replacement of %20 with +. */` I didn't understand this. Also there is */ at the end of it

tmansuroglu: lines below made it clear but I should be able to understand it in the comments

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+  const formEncodeString = (str) =>
+    encodeURIComponent(str).replace(/%20/g, "+");
+
+  // Converts object to proper form data
+  // name/value pairs are separated by the ampersand (&)
+  // names are separated from values by the equals symbol (=).
+  const encodeData = (obj) =>
+    Object.entries(obj)
+      .map(
+        (pair) => formEncodeString(pair[0]) + "=" + formEncodeString(pair[1])
```
tmansuroglu, line 35: pair[0] and pair[1] should be stored in a const with appropriate name

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+        "form-name": formName,
+        info: honey,
+      }),
+    });
+  };
+
+  // checks beforeSubmit, onsubmit, afterSubmit works and has errors
+  const onSubmit = async (e) => {
+    e.preventDefault();
+    if ((beforeSubmit && (await beforeSubmit())) || !beforeSubmit) {
```
tmansuroglu, line 58: Why would beforeSubmit not exist or be false? What does beforeSbumit do?

tmansuroglu: Same questions apply to other variables here

src/components/ContactUsForm/index.css
```diff
@@ -0,0 +1,58 @@
...
+  box-sizing: border-box;
+  border: 2px solid #ccc;
+  border-radius: 0.5rem;
+  font-size: 16px;
+  width: clamp(20vw, 30vw, 40vw);
+  height: clamp(3vh, 5vh, 6vh);
+}
+.contactUsForm__content__textarea {
+  height: clamp(8vh, 10vh, 16vh);
+}
```
tmansuroglu: Louis warned us about use of viewport

tmansuroglu (changes requested):

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks Tarkan, I'll make the changes that you've mentioned.

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <div className="contactUsForm">
+      <NetlifyForm
+        formName="contact"
+        contactFormValues={contactFormValues}
+        beforeSubmit={beforeSubmit}
+        afterSubmit={afterSubmit}
+        automaticHoneypot={true}
+      >
+        <span className="contactUsForm__result">{submissionMessage}</span>
```
YamanSammani, line 58: Thanks to Tarkan for taking care of most of the comments, I have a small one here, 

Do you think we need to consider using conditional rendering of this so it doesn't render an empty span when submittionMessage is empty?

src/components/NetlifyForm/index.jsx
ssinikoglu: > Why would beforeSubmit not exist or be false? What does beforeSbumit do?

If beforeSubmit is true, it will not validate form and button will keep its initial state(disabled).

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for your review Yaman, I used span, because it's inline- block element. But I'll try to add conditional rendering to the span as you've mentioned.

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for your review Tarkan, I  have changed it.

src/components/ContactUsForm/index.jsx
ssinikoglu: I've changed it as well.

src/components/ContactUsForm/index.jsx
ssinikoglu: All class names are updated now.

src/components/NetlifyForm/index.jsx
ssinikoglu: Thanks Tarkan, I wrote inline notes for them.

src/components/NetlifyForm/index.jsx
ssinikoglu: It keeps hidden Honey form field, I also updated my comment so that it could understand better.

src/components/NetlifyForm/index.jsx
ssinikoglu: I added more detailed explanation for honeypot form field and I also shared the documentation link that I've read.

src/components/NetlifyForm/index.jsx
ssinikoglu: The sentence was not clean enough, so I changed it and also add documentation link.

src/components/ContactUsForm/index.css
ssinikoglu: Thanks for reminding;) I updated all viewport scales into rem.

src/components/ContactUsForm/index.jsx
ssinikoglu: I've changed it as well.

package.json
```diff
@@ -11,13 +11,15 @@
         "antd": "^4.7.2",
         "firebase": "^7.24.0",
         "firebaseui": "^4.7.0",
+        "gatsby-link": "^2.4.16",
```
louisrli, line 14: you should not need gatsby-link . you're not using gatsby in this pr

src/components/ContactUsForm/index.css
```diff
@@ -0,0 +1,60 @@
+.contactUsForm {
+background: #add4c5;
```
louisrli, line 2: indentation is all wrong

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,94 @@
+/*
+ * NetlifyForm component consists of server related functionalities that is used posting data and also avoiding captchas.
+ * The base code that is used in both netlifyForm and index are implemented from https://github.com/jon-fm/react-ssg-netlify-forms-demo
+ */
+import React, { useState, useEffect } from "react";
+
+const NetlifyForm = ({
+  formName,
+  beforeSubmit, // function which is used in form validation, disabling submit button
```
louisrli, line 9: comments should go above line 7

src/components/NetlifyForm/index.jsx
```diff
@@ -0,0 +1,94 @@
+/*
+ * NetlifyForm component consists of server related functionalities that is used posting data and also avoiding captchas.
+ * The base code that is used in both netlifyForm and index are implemented from https://github.com/jon-fm/react-ssg-netlify-forms-demo
```
louisrli, line 3: so i hate to tell you this but this is why we dont copy code with 0 stars from github. its waaaaay unnecessarily complicated for what you're trying to do

also that demo is for "react-ssg" which if you google it is not what you're doing

you can still use netlify forms but just follow the tutorial here, and change ContactForm

what should have been done is to follow this tutorial here:
https://docs.netlify.com/forms/setup/#submit-javascript-rendered-forms-with-ajax

which i think would have been a lot easier overall

src/components/NetlifyForm/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,17 @@
...
+
+exports[`ContactUsForm snapshot test 1`] = `
+<form
+  data-netlify="true"
+  netlify-honeypot="info"
+  onSubmit={[Function]}
+/>
+`;
+
+exports[`NetlifyForm snapshot test 1`] = `
```
louisrli, line 11: delete one of these, as you can see both of these snapshots show the exact same thing so its not very useful to have two

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+  const [processing, setProcessing] = useState(false);
+  // beforeSubmit is for validations and disabling submit button
+  const beforeSubmit = async () => {
+    if (
+      contactFormValues.name.length > 0 &&
+      contactFormValues.email.length > 0
+    ) {
+      setProcessing(true);
+      // Wait 2 seconds to simulate async delay
+      await new Promise((resolve) => setTimeout(resolve, 2000));
```
louisrli, line 25: why do you need to simulate async delay? this seems wrong

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+      [e.target.name]: e.target.value,
+    });
+  const [contactFormValues, setContactFormValues] = useState({
+    name: "",
+    email: "",
+    message: "",
+  });
+  return (
+    <div className="contactUsForm">
+      <NetlifyForm
```
louisrli, line 51: i think if you delete this whole NetlifyForm thing and just add data-netlify=true to the form (https://docs.netlify.com/forms/setup/#html-forms) then it will refresh the page but also submit the data. did you try that?

because that would reduce submitting to netlify forms to one line rather than having this complicated component/onsubmit thing (see my other comment about copying a random person's code)

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+    ) {
+      setProcessing(true);
+      // Wait 2 seconds to simulate async delay
+      await new Promise((resolve) => setTimeout(resolve, 2000));
+      return true;
+    } else {
+      return false;
+    }
+  };
+  // TODO: get submission message to translate files.
```
louisrli, line 31: can you just do this in this PR since you added the translations anyway?

louisrli:
can you please try to follow the official documentation here and delete NetlifyComponent and see if that gives you trouble. it will simplify the code considerably
https://docs.netlify.com/forms/setup/#html-forms

try submit forms with html first, then try with ajax second (the first one will cause page refresh, the second won't)

usually we only want this NetlifyForm type thing if we are planning to have a lot of forms in which we will reuse its logic, but as far as i know you just have the contactform on this site

louisrli:
if it doesn't work, just roll back the commit and keep it this way, but i don't see why the official documentation wouldn't work

src/components/NetlifyForm/index.jsx
tmansuroglu: I will take a second look at this. I might be mistaken

PR #55, opened 2020-11-15
Memory game board

tmansuroglu:
#22 I have refactored final code to make it more readable. This is a part of it. So, majority of the code is already reviewed. Also, we agreed with Ammar to not add tests. Something is wrong and Ammar will fix it when all code is merged. Tests will be added during bug bash week.

### DIFF
--- a/src/components/QuizGameApi/index.jsx
+++ b/src/components/QuizGameApi/index.jsx
@@ -1,0 +1,16 @@
+// we fetch those questions from opentdb website. we can pick the type of questions,the amount and the dificulty.
+// here we have a 10 questions as an amount, nature as a category ,easy, middle and hard for the difficulty and multiple choices as a type.
+// iam fetching data from a website called opentdb.com this website will give us a simple questions about the nature.
+// this is an example of the game https://lovattspuzzles.com/online-puzzles-competitions/ultimate-online-trivia-quiz.
+import shuffle from "../../components/memorygame/shuffle.jsx";
+
+const NATURE_CATEGORY = 17;
+
+export const fetchQuizQuestions = async (amount, difficulty) => {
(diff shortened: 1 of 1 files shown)


src/components/memorygame/MemoryGameBoard/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      <Row align="middle" className="howToPlayPopoverAndScrollToFactsRow">
+        <Col span={5}>
+          <ScrollToFacts />
+        </Col>
+        <Col span={2} push={17}>
+          <HowToPlayPopover />
+        </Col>
+      </Row>
+      <Row>
+        <GameCards
```
louisrli: unless you completely rewrote the code, notice how the diff on the left side is empty and its saying you added 128 lines.

you want to use git mv to move MemoryGameCards/idnex.js to MemoryGameBoard/index.js, then modify it.

imagine you are me. i am wondering whether the code was completely rewritten or whether you rewrote just parts of it. but now i have to rereview the whole thing.

https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move

basically you shouldnt be deleting and then copy-pasting to a new file, you should be doing `git mv` followed by `git commit` and that should make the diff show up correctly (unless github is strange about it). but give it a try.

louisrli: what i should see is a rename + a diff of how the code was changed. otherwise i'm rereviewing the whole code.

tmansuroglu:
I closed this by mistake while deleting extra commits

majdajroudi:
is this a PR that we still have to review? Files changed tab is empty

tmansuroglu:
No, I messed this up. There is a open version of this.

PR #56, opened 2020-11-15
memory game fact list refactored

tmansuroglu:
#22 Part of refactored final code.


src/components/memorygame/MemoryGameFactList/index.jsx
```diff
@@ -1,31 +1,65 @@
...
+ * If the language is English "emptyFactListMessage" will be ""Match cards to reveal facts!"
  */
-const MemoryGameFactList = ({ facts, title }) => {
-  return (
-    <>
-      <h1 className="memoryGame__title">{title}</h1>
+const MemoryGameFactList = ({ facts }) => {
+  // facts example:
+  // {
+  //   cardKey: "alternateEnergy0",
```
louisrli: again i dont recommend doing soemthing like this in futrue projects but its fine for now. generally you'd want to use something like typescript to define an interface. the problem with this approach is that its very brittle to code changes, there is nothing stopping you from adding/removnig a field from facts and forgetting that this comment exists.

src/components/memorygame/MemoryGameFactList/index.jsx
```diff
@@ -1,31 +1,65 @@
...
+  //   description: "Alternatif Enerji",
+  //   img: {src: "/static/media/alternateEnergy.88cef00e.svg", imgKey: "alternateEnergy"},
+  //   isFlipped: false,
+  //   isMatched: false,
+  //   link: "https://loremIpsum.com",
+  //   phrase: "Lorem ipsum",
+  // }
+  const [t] = useTranslation();
+  const factsTitle = t("memoryGame.facts");
+  const emptyFactListMessage = t("memoryGame.emptyFactListMessage");
```
louisrli: i would just inline lines 27 to 31. a bit too overboard with the variables, they are perfectly readable if they are inline / they are not being used more than once. i'd say the code is catually harder to read when it's here

src/components/memorygame/MemoryGameFactList/index.jsx
```diff
@@ -1,31 +1,65 @@
...
+  const factListTitle = (
+    <h1 className="memoryGameFactListContainer__title">{factsTitle}</h1>
+  );
+  let factListContent;
+  // facts is an empty array by default, it gets filled when user makes successful match
+  if (facts.length) {
```
louisrli: facts.length !== 0, be explicit, you may think it's shorter and fancy but it takes the reader a second to think about this

louisrli (changes requested):

src/components/memorygame/MemoryGameFactList/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFactList/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFactList/index.jsx
tmansuroglu: PTAL

louisrli (approved):

ssinikoglu (approved):
Nice description and code, I don't have any comments to add.

ibrahimaljarad (approved):

tmansuroglu:
@Ammar-64  can you merge?

PR #57, opened 2020-11-15
memory game flippable card

tmansuroglu:
#22


src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+// check https://github.com/AaronCCWong/react-card-flip for 'react-card-flip' details
+import ReactCardFlip from "react-card-flip";
+import { Card } from "antd";
+import questionMark from "../../../images/memoryGameFrontImage.png";
+import "./index.css";
+
+// targetState example
+// {
+//   cardKey: "alternateEnergy0",
```
louisrli: now that this comment has been copied in multiple places that would not be acceptable as committed code, because it means that if i add a field to this state i need to add it to at least 2+ places (i think this is the same as the other comment in the other PR right?)

what you probably want to do is put a comment at the top of the main memory game component that says this. here, all you need to write is "see MemoryGameBoard/index.js for an example of what this object looks like" (it's a 'pointer' to that documentation)

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+//   isMatched: false,
+//   link: "https://lorem.com",
+//   phrase: "Lorem ipsum"
+// }
+// FlippableCard is being used in a map function. So targetState refers to each iteration of map
+// Mapped array is cardStates. it holds states like example above(line 9-17)
+// selectedCards is an array of clicked card states(objects).
+// selectedCards array hold only "NUM_REQUIRED_MATCHES" amount of card states
+// example of selectedCards is same as example above(line 9-17)
+
```
louisrli: extra newline

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+//   img: {src: "/static/media/alternateEnergy.88cef00e.svg", imgKey: "alternateEnergy"},
+//   isFlipped: false,
+//   isMatched: false,
+//   link: "https://lorem.com",
+//   phrase: "Lorem ipsum"
+// }
+// FlippableCard is being used in a map function. So targetState refers to each iteration of map
+// Mapped array is cardStates. it holds states like example above(line 9-17)
+// selectedCards is an array of clicked card states(objects).
+// selectedCards array hold only "NUM_REQUIRED_MATCHES" amount of card states
```
louisrli: rather than writing this comment its better to write an invariant inside the function body that this is true, and you can delete this comment. let your code enforce this fact

but also i think this comment might be wrong, can't selectedCards be less than NUM_REQUIRED_MATCHES? you can have only one card selected. in that case, these comments or variable names need to be fixed

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+    // current index is needed to keep the order of cards displayed
+    // because goal is to make user remember the spots of previous cards
+    const currentIndex = cardStates.findIndex(
+      (state) => state.cardKey === clickedCardState.cardKey
+    );
+    const targetCardState = copyOfCardStates[currentIndex];
+    targetCardState.isFlipped = true;
+    setCardStates(() => [...copyOfCardStates]);
+    // prevents user to use same card multiple times
+    const isCardNotInSelectedCards = selectedCards.every(
```
louisrli: do not use Not in your variable names for booleans

try reading this:
`!isCardNotInSelectedCards`

read code complete section about variable naming; prefer positive, intuitive names for your variables
"isCardInSelectedCards"
"!isCardInSelectedCards"

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  selectedCards,
+  setSelectedCards,
+}) => {
+  // handle click flips card
+  // if the card is not in selectedCards, adds it to selectedCards
+  const handleClick = (clickedCardState) => {
+    const copyOfCardStates = [...cardStates];
+    // current index is needed to keep the order of cards displayed
+    // because goal is to make user remember the spots of previous cards
+    const currentIndex = cardStates.findIndex(
```
louisrli: clickedCardIndex

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+    // prevents user to use same card multiple times
+    const isCardNotInSelectedCards = selectedCards.every(
+      (selectedCard) => selectedCard.cardKey !== clickedCardState.cardKey
+    );
+    if (isCardNotInSelectedCards) {
+      setSelectedCards((prevState) => [...prevState, clickedCardState]);
+    }
+  };
+  return (
+    // ReactCardFlip requires 2 children(front side and back side) and isFlipped value to work
```
louisrli: you can delete "and isFlippedvalue to work" in this comment, only the first part of it is something that might be confusing. isflipped already speaks for itself

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: Yes comment is wrong. It can hold less than "NUM_REQUIRED_MATCHES". What I meant was useEffect returns nothing if its length is less than "NUM_REQUIRED_MATCHES". I will fix it

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: I have deleted it. invariant will be added to MemoryGameBoard component. PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: PTAL

src/components/memorygame/MemoryGameFlippableCard/index.css
```diff
@@ -0,0 +1,18 @@
+.memoryGameCardsContainer___card {
+  cursor: pointer;
+  height: clamp(20vh, 280px, 35vh);
+  width: 100%;
+  border-radius: 20px;
+  transition: all, 0.5s;
+}
+
+.memoryGameCardsContainer___card___image {
```
louisrli, line 9: imo something in its own component shouldn't have to know that it belongs to memoryGameCardsContainer. that violates the whole point of the component being independent from its parent components. just give it a css class that is scoped to the card

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+    // current index is needed to keep the order of cards displayed
+    // because goal is to make user remember the spots of previous cards
+    const clickedCardIndex = cardStates.findIndex(
+      (state) => state.cardKey === clickedCardState.cardKey
+    );
+    const targetCardState = copyOfCardStates[clickedCardIndex];
+    targetCardState.isFlipped = true;
+    setCardStates(() => [...copyOfCardStates]);
+    // prevents user to use same card multiple times
+    const isCardInSelectedCards = selectedCards.every(
```
louisrli: shouldn't this be some, not every?

louisrli (changes requested):

src/components/memorygame/MemoryGameFlippableCard/index.css
louisrli: nvm ignore this comment. I thought that memorygamecardscontainer was the parent css class but I believe it's only used here, which is good

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: Correct it should be some. PTAL

src/components/memorygame/MemoryGameFlippableCard/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+    const copyOfCardStates = [...cardStates];
+    // current index is needed to keep the order of cards displayed
+    // because goal is to make user remember the spots of previous cards
+    const clickedCardIndex = cardStates.findIndex(
+      (state) => state.cardKey === clickedCardState.cardKey
+    );
+    const targetCardState = copyOfCardStates[clickedCardIndex];
+    targetCardState.isFlipped = true;
+    setCardStates(() => [...copyOfCardStates]);
+    // prevents user to use same card multiple times
```
louisrli, line 30: does your game handle unflipping if a card is selected?

i pick card 0

then i click card 0 again (its the first selected card)

does it unflip?

you don't have to do this, i'm just asking, but then you also probably want flipped cards to have a not-allowed over it if you didnt already

louisrli (approved):

src/components/memorygame/MemoryGameFlippableCard/index.jsx
ssinikoglu: I thought it is for keeping correctly matching cards, if it is true match user can't unflip.

ibrahimaljarad (approved):

src/components/memorygame/MemoryGameFlippableCard/index.jsx
tmansuroglu: Unflipping is done automatically when user makes a wrong match. So if you select same card twice it doesnt unflip. I added conditional class selection to implement not allowed cursor. PTAL

tmansuroglu:
@Ammar-64  can you merge?

tmansuroglu:
@Ammar-64  up to date and ready for merge

PR #60, opened 2020-11-15
memory game changeCardStatePropToOpposite function

tmansuroglu:
#22


src/components/memorygame/changeCardStatePropertyToOpposite.jsx
```diff
@@ -0,0 +1,48 @@
+// manipulates isFlipped and isMatched property of card a state
```
louisrli: you dont want to put functions one per file like this. memorygame/gameHelpers.jsx

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
```diff
@@ -0,0 +1,48 @@
...
+// property example : "isMatched"
+// cardStates is an array of objects(states)
+// cardStates example :
+// [
+//   {
+//      cardKey:"",
```
louisrli: same comment here as your other prs

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
```diff
@@ -0,0 +1,48 @@
...
+//      img:{src:"",imgKey:""},
+//      isFlipped:false,
+//      isMatched:false,
+//      description:"",
+//      link:"",
+//      phrase:""
+//   },
+// ...]
+// targetCard is a state obj.
+// Object in line 6-14 is an example for targetCard
```
louisrli: using line numbers in comments guarantees that your comment will be useless when someone else that's not you edits this file, unless you are expecting every teammate to read every comment and look for things like this

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
```diff
@@ -0,0 +1,48 @@
...
+const changeCardStatePropertyToOpposite = (
+  property,
```
majdajroudi: Where is the property prop coming from? I got a bit confused with the if conditions

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
```diff
@@ -0,0 +1,48 @@
...
+//      isFlipped:false,
+//      isMatched:false,
+//      description:"",
+//      link:"",
+//      phrase:""
+//   },
+// ...]
+// targetCard is a state obj.
+// Object in line 6-14 is an example for targetCard
+const changeCardStatePropertyToOpposite = (
```
majdajroudi: is this fn called on click of the card?

majdajroudi (commented):
generally readable but needs more comments at the top to explain where the function is called

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
tmansuroglu: Since this is a function, property is not a prop. It is a parameter. Name is misleading but function changes property of a state.So, property stands for target state property. I added a comment about where to see the use of function. Also, I have changed property into stateProperty. I'm not sure if it is descriptive enough. Please let me know

PTAL

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
tmansuroglu: PTAL

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
tmansuroglu: PTAL

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
tmansuroglu: PTAL

src/components/memorygame/changeCardStatePropertyToOpposite.jsx
tmansuroglu: added comment PTAL

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,46 @@
+// manipulates isFlipped and isMatched property of card a state
+// property example : "isMatched"
+// cardStates is an array of objects(states)
+// see src/memorygame/MemoryGameBoard/index.jsx to for cardStates example and use of this function
+// targetCard is an element of cardStates array.
+// changeCardStatePropertyToOpposite is called when selectedCards state change.
+export const changeCardStatePropertyToOpposite = (
```
louisrli: invertCardStateForKey 

we call them keys usually. bc objects are key value pairs

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,46 @@
...
+  stateProperty,
```
louisrli: cardstatekey

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,46 @@
...
+  const copyOfCardStates = [...cardStates];
+  const targetIndex = copyOfCardStates.findIndex(
+    (cardState) => cardState.cardKey === targetCard.cardKey
+  );
+  const targetCardState = copyOfCardStates[targetIndex];
+  // isMatched is false as default. When it is true success symbol appears in card
+  if (stateProperty === "isMatched") {
+    targetCardState.isMatched = !targetCardState.isMatched;
+    setCardStates(() => [...copyOfCardStates]);
+    // isFlipped is false as default. When it is true backside of card become visible
```
louisrli: don't put comments above else like this. it's confusing because it's in the if condition above it. put it below the else

louisrli: you can just delete this comment I think. no need to describe isflipped here, islfipped is probably the most intuitive variable in the code

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,46 @@
...
+  } else {
+    throw new Error(
+      "Invalid property was passed to changeCardStatePropertyToOpposite"
+    );
+  }
+};
+
+// Check Fisher–Yates shuffle algorithm https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
+// non destructive shuffle
+export const shuffle = (arr) => {
```
louisrli: uhh aren't your teammates supposed to be using the same shuffle? what happened to that

louisrli (changes requested):

src/components/memorygame/gameHelpers.jsx
tmansuroglu: makes sense PTAL

src/components/memorygame/gameHelpers.jsx
tmansuroglu: PTAL

src/components/memorygame/gameHelpers.jsx
tmansuroglu: comment deleted PTAL

src/components/memorygame/gameHelpers.jsx
tmansuroglu: Correct. shuffle moved back to its previous file. PTAL

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,34 @@
+// manipulates isFlipped and isMatched property of card a state
+// property example : "isMatched"
```
majdajroudi: you should change this comment to `cardStateKey example: ...`

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,34 @@
...
+// cardStates is an array of objects(states)
+// see src/memorygame/MemoryGameBoard/index.jsx to for cardStates example and use of this function
+// targetCard is an element of cardStates array.
+// changeCardStatePropertyToOpposite is called when selectedCards state change.
```
majdajroudi: `invertCardStateForKey is called ...`

src/components/memorygame/gameHelpers.jsx
```diff
@@ -0,0 +1,34 @@
...
+  } else if (cardStateKey === "isFlipped") {
+    // at this point isFlipped is always true and user failed to match cards.
+    // resets it back to false. But gives player time to see failed cards
+    setTimeout(() => {
+      targetCardState.isFlipped = !targetCardState.isFlipped;
+      setCardStates(() => [...copyOfCardStates]);
+    }, 750);
+  } else {
+    throw new Error(
+      "Invalid property was passed to changeCardStatePropertyToOpposite"
```
majdajroudi: `... passed to invertCardStateForKey`

majdajroudi (changes requested):

majdajroudi:
Generally good but comments need to be updated

src/components/memorygame/gameHelpers.jsx
louisrli: majd has some really good points here in general tarkan, which for both of you, the general principle will be a good takeaway. it's good that both of you are writing more comments and developing an intuition for whats readable. the next step is also to write "maintainable" comments

even a small change requested can cause your comments be outdated. and we (we as in we programmers) dont have time to remember comment dependencies (even in a typed language like typescript, this is not "typed" properly in comments, so your code will still work even if comments are outdated)

the lesson / goal here is to write your comments to be as futureproof as possible (its not completely possible and you dont want to be vague but its a goal to aim for). for example, avoiding line numbers. or, when you're naming props, you can put the comment above the prop for example without mentioniing its name (in JS theres not a good way to do this). thus even if you rename the prop the comment is still above the prop

louisrli (commented):
when majd approves i will approve

src/components/memorygame/gameHelpers.jsx
tmansuroglu: By saying `you can put the comment above the prop for example without mentioniing its name` are you referring to something like below? @louisrli 
```
function(
  // does bla bla
  param1,
  // does different bla
  param2
)
```

src/components/memorygame/gameHelpers.jsx
tmansuroglu: PTAL

src/components/memorygame/gameHelpers.jsx
tmansuroglu: PTAL

majdajroudi (approved):

louisrli:
yes

louisrli:
to the comments thing

tmansuroglu:
@louisrli majd approved it

tmansuroglu:
@louisrli waiting for approval

louisrli (approved):

tmansuroglu:
@Ammar-64 ready for merge

PR #61, opened 2020-11-15
memory game container and confetti config files

tmansuroglu:
#22 This should be merged after memory game board


ibrahimaljarad (approved):
i see it is a small readable component and i am ok with everything

src/containers/MemoryGame/index.jsx
```diff
@@ -1,12 +1,44 @@
...
+import MemoryGameBoard from "../../components/memorygame/MemoryGameBoard";
 
 const MemoryGame = () => {
+  const [t] = useTranslation();
+  const [isGameCompleted, setIsGameCompleted] = useState(false);
+  // matchedCards is being used to display environmental facts.
+  // It is an array of object(states).Each object is a successfully matched card state.
+  // When a match is made only state of first card gets pushed inside matchedCards.
+  // Because matched cards have same state. Otherwise same fact would be displayed twice.
+  const [matchedCards, setMatchedCards] = useState([]);
```
louisrli: does matched here mean completed like it can no longer be flipped? I'd call this finished cards if that's the case. otherwise the word match is ambiguous

louisrli: since the whole game is about matching it is also matched cards when I try to match one card for acid rain and one card for oceans or something. that would also be considered matching

louisrli (changes requested):

src/containers/MemoryGame/index.jsx
tmansuroglu: Yes matched means it can no longer be flipped. Changes made PTAL

src/containers/MemoryGame/index.jsx
```diff
@@ -1,12 +1,44 @@
...
+import MemoryGameBoard from "../../components/memorygame/MemoryGameBoard";
 
 const MemoryGame = () => {
+  const [t] = useTranslation();
+  const [isGameCompleted, setIsGameCompleted] = useState(false);
+  // finishedCards is being used to display environmental facts.
+  // It is an array of object(states).Each object is a successfully matched card state.
+  // When a match is made successfully only state of first card gets pushed inside finishedCards.
+  // Because matched cards have same state. Otherwise same fact would be displayed twice.
+  const [finishedCards, setFinishedCards] = useState([]);
```
louisrli, line 17: you dont need to change this here (please don't in any of the PRs you sent out because i don't want to rereview the PRs, do it in a separate PR if you want) but unless you need the whole state, the cleanest way to implement this would not be to push around card states, but probably to only store the imageKey of finished cards. then using the image key you would look up the facts related to that image

basically you can think that in your current structure of passing around data you are duplicating/denormalizing data a lot, which is risky esp. in a language like javascript where you are mutating some parts of the state etc.

in general a warning sign would be any code that says "i duplicated these items but i only need to keep one of them, but you know, all of them are the same." (except that a HUGE class of bugs is going to be when the duplicates actually differ in some way you didn't expect). storing only the image key and only looking up the facts by the image key guarantees that this is not possible.

finished cards -> store image key -> look up facts by image key for finished fact list

src/containers/MemoryGame/index.jsx
```diff
@@ -1,12 +1,44 @@
...
+          <MemoryGameBoard
+            setFinishedCards={setFinishedCards}
+            finishedCards={finishedCards}
+            setIsGameCompleted={setIsGameCompleted}
+          />
+        </Row>
+      </Col>
+      <Col xs={20}>
+        <Confetti active={isGameCompleted} config={confettiConfig} />
+        <MemoryGameFactList facts={finishedCards} />
```
louisrli, line 41: related to the last comment, as a general rule, components should only know what they need to. if i had to write tests for MemoryGameFactList, i'd have to write a bunch of things about cards being flipped and so on. and also i'd have to remember that it should only take de-deduplicated card states

this would be an example of an unclean interface. components should be "dumb" and well-defined. memorygamefactlist should be "dumb" in the sense that it should say: "i don't know anything about card states. i don't know anything about flipping. all i know is that you're gonna give me some facts and i'm going to show them."

also the fact that it has a prop called `facts` that takes card states already seems wrong (the interface is good in spirit but facts !== cardState in terms of "types")

louisrli (approved):

ssinikoglu (approved):

src/containers/MemoryGame/index.jsx
tmansuroglu: I noted this. After finishing my issues I will try to help others. If we still have time after that I will implement this.

src/containers/MemoryGame/index.jsx
tmansuroglu: Understood.

tmansuroglu:
@Ammar-64 can you merge?

PR #63, opened 2020-11-16
ContactUsForm Component #35

### LINKED ISSUE #35: Contact page

ssinikoglu:
This PR consists of ContactUsForm component of Contact page. I created new PR since I changed all the format and documentation of the contact form (NetlifyForm file has been deleted now.). The PR is written according to via [this](https://www.netlify.com/blog/2017/07/20/how-to-integrate-netlifys-form-handling-in-a-react-app/) Netlify documentation. 

Upcoming PRs for the Contact Page will be ContactUsBottom (the component has only visual features (curve and images) and responsive design.), Contact Container.

Please review. I am looking forward to your comments.


src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+    name: "",
+    email: "",
+    message: "",
+  });
+
+  // The change in isSubmitted state triggers submission message
+  const [isSubmitted, setIsSubmitted] = useState(false);
+
+  // Converts object to proper form data(string)
+  const encode = (data) => {
```
louisrli: since this is a "pure function" (can google it) that is independent from the component i'd say it actually makes more sense to move this below the imports (in the same file), but outside the component. and call it "encodeFormData"

louisrli: separating functions like that makes it clear to the reader that it has no dependency on anything inside the component.

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  // encode is used for passing the string to the body
+  // setTimeOut stands for showing submission message about a time after submit
+  const handleSubmit = (e) => {
+    fetch("/", {
+      method: "POST",
+      "data-netlify": "true",
+      headers: { "Content-Type": "application/x-www-form-urlencoded" },
+      body: encode({ "form-name": "contact", ...formValue }),
+    }).then(() => {
+      setTimeout(() => setIsSubmitted(false), 5000);
```
louisrli: this timeout is weird. you probably don't want to do this, and 5 seconds is a long time for the user to wait. you may want a state called isLoading or something (ibrahim did something similar) but in general we never write code with manual timeouts like this. it could take, in reality, longer than 5 seconds.

what's the reason?

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+      setIsSubmitted(true);
+      setFormValue({
+        name: "",
+        email: "",
+        message: "",
+      }).catch((error) => console.log(error));
```
louisrli: generally we never console.log, prefer to alert(error) here if you want a low fidelity error, otherwise
```
const [submitError, setSubmitError] = React.useState(null)

.catch((error) => setSubmitError(error))

<div class="errorMessageDiv">{submitError}</div>
```

something like that

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  return (
+    <div className="contactUsFormContainer">
+      <form onSubmit={handleSubmit}>
+        <Col>
+          <div
+            className="contactUsForm__submissionMessage"
+            style={
+              isSubmitted ? { visibility: "visible" } : { visibility: "hidden" }
```
louisrli: style={{visibility: isSubmitted ? "visible" : ... }}

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+            }
+          >
+            <strong>{t("contact.contactUsForm.submissionMessage")}</strong>
```
louisrli: i would call this contactUsForm.postSubmissionMessage to be super clear

louisrli: post meaning "after"

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+              onChange={handleChange}
+              className="contactUsForm__textarea"
+              required
+            />
+          </Col>
+        </Col>
+        <Row>
+          <div className="contactUsForm__submit">
+            <button type="submit" className="contactUsForm__submitButton">
+              {" "}
```
louisrli: this line should be deletable

louisrli (commented):
mostly looks good actually. nice work following the netlify docs. some minor comments

src/components/ContactUsForm/index.jsx
ssinikoglu: I removed the line, PTAL.

src/components/ContactUsForm/index.jsx
ssinikoglu: All submissionMessage related names are changed, PTAL.

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for your review, it's format changed and become cleaner, PTAL.

src/components/ContactUsForm/index.jsx
ssinikoglu: I changed console.log into alert for now.

src/components/ContactUsForm/index.jsx
ssinikoglu: The main reason I have used timeout (via using isSubmitted state) is that showing post submit message a little more to the user. I gave 5 seconds by trying and it seems I underestimated back sides of it. I'm looking Ibrahim's code, I'll try adapting it to  handleSubmit function.

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for reviewing and explanation. I moved the function to the new file. PTAL

package.json
```diff
@@ -7,7 +7,6 @@
         "@fortawesome/fontawesome-svg-core": "^1.2.32",
         "@fortawesome/free-brands-svg-icons": "^5.15.1",
         "@fortawesome/react-fontawesome": "^0.1.12",
-        "@ant-design/icons": "^4.2.2",
```
louisrli: is this intended to be deleted?

src/components/ContactUsForm/encodeFormData.jsx
```diff
@@ -0,0 +1,6 @@
+// Converts object to proper form data(string)
+export const encodeFormData = (data) => {
```
louisrli: actually it's cleaner to put this in the same file as.the component. under the imports. it's harder for readability when one function is in its own file because people need to jump files to read it

louisrli (changes requested):
looks good. will approve when tests pass

package.json
ssinikoglu: It is not comletely deleted, version of the icons changed after latest merge.

src/components/ContactUsForm/encodeFormData.jsx
ssinikoglu: I moved the function under the component file. PTAL

ssinikoglu:
> looks good. will approve when tests pass

@majdajroudi fixed the problem, all tests are passing now.

louisrli (approved):

src/components/ContactUsForm/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+        email: "",
+        message: "",
+      }).catch((error) => alert(error));
+
+      e.preventDefault();
+    });
+  };
+
+  return (
+    <div className="contactUsFormContainer">
```
tmansuroglu, line 51: if you change className to "contactUsForm" all class names will be correct

tmansuroglu (approved):
Nice work!

ssinikoglu:
Hi @Ammar-64, PR is approved, PTAL.

PR #66, opened 2020-11-17
About page - Main component

majdajroudi:
This PR contains the about page container. I added the illustrations and the main text of the page to this PR, I will create one more PR for the team members' cards and then integrate both components with one final PR when they are both approved. 

aboutPage


src/containers/AboutPage/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+import "./index.css";
+
+const AboutPage = () => {
+  const [t] = useTranslation();
+
+  return (
+    <div className="aboutUs">
+      <Row className="aboutUs__upperContent">
+        <Col className="aboutUs__textCol">
+          <Row className="aboutUs__textCol--aboutText">
```
ssinikoglu, line 15: For the styling of About us paragraph, you may add `text-align: justify;` it's a suggestion because I'm not sure how this styling will look in Arabic version. You may consider it also for the Why do we do it paragraph. Other than this, you may add font-styling (font-family, size, color) for the content.

ssinikoglu (commented):
Class names, code are easy to understand and well-organized.  I added  a small comment about styling.

src/containers/AboutPage/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+            <h2>{t("about.aboutUs")}</h2>
+            <p>{t("about.aboutUsText")}</p>
+          </Row>
+          <Row className="aboutUs__textCol--whyWeDoIt">
+            <h2>{t("about.whyWeDoIt")}</h2>
```
louisrli: whyWeDoItHeader

louisrli: same with other headers

src/containers/AboutPage/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+            <p>{t("about.whyWeDoItText")}</p>
+          </Row>
+        </Col>
+        <Col className="aboutUs__illustrationsCol">
+          <img
+            className="aboutUs__illustrationsCol--kid1"
+            src={kid1Img}
```
louisrli: kid0 kid1

louisrli (approved):

ssinikoglu (approved):

src/containers/AboutPage/index.jsx
majdajroudi: I can't believe I'm still making this mistake, PTAL

src/containers/AboutPage/index.jsx
majdajroudi: PTAL

src/containers/AboutPage/index.jsx
majdajroudi: I used the `text-align` as suggested, and I used the the "Salsa" font for the headers but kept the paragraph's text as default because it's important to be readable, as suggested by Berfin. regarding the colors and the size I didn't modify them because as we agreed in the last standup meeting we need to discuss those and agree on one color for the context's text and their readability. 

PTAL

ssinikoglu: You are exactly right, we should decide and go with using one of them.

majdajroudi:
PR approved, PTAL @Ammar-64

Issue #34, opened 2020-11-09
Games Page 


ssinikoglu:
Make games page design similar to figma, connect it to single games, create required tests, make it responsive, set up nested routing.

PR #75, opened 2020-11-18
Game Page Image grid rework

tmansuroglu:
#34 Previous layout didn't work as I intended. I put cards instead of divs with background image


ibrahimaljarad (approved):
nice and clean,

src/components/gamespage/GamesPageImageGrid/index.css
```diff
@@ -1,49 +1,40 @@
...
-  background: rgb(251, 250, 255);
-  background: linear-gradient(
-    0deg,
-    rgba(251, 250, 255, 1) 0%,
-    rgba(0, 0, 0, 0.2553396358543417) 0%,
-    rgba(0, 0, 0, 0.9164040616246498) 0%,
-    rgba(0, 0, 0, 0.5242471988795518) 0%
-  );
+  font-size: clamp(0.9rem, 2.7vw, 3rem);
+  font-family: "Salsa";
```
louisrli: why does the other one have cursive but this doesn't

src/components/gamespage/GamesPageImageGrid/index.css
```diff
@@ -1,49 +1,40 @@
...
-.gamePageImageGrid__col:hover {
-  transform: scale(1.1);
+.gamePageImageGrid__col__link__card:hover {
+  transform: scale(1.05);
   transition: all, 0.5s;
 }
 
-.gamePageImageGrid__col__redirect {
-  display: inline-flex;
-  flex-direction: column;
```
louisrli, line 18: why are the tests deleted

ssinikoglu (approved):
Overall, it looks nice.

louisrli (changes requested):

src/components/gamespage/GamesPageImageGrid/index.css
tmansuroglu: Because `gameData `prop is `undefined` and test throws error. It will be defined when component merges with container.

src/components/gamespage/GamesPageImageGrid/index.css
tmansuroglu: cursive added PTAL

src/components/gamespage/GamesPageImageGrid/index.css
louisrli: you don't delete tests when they don't work. you change the way they're called. also you should be passing in mock game data anyway in the test...

louisrli (commented):
the point of having a component called `<GamePageData gameData=` that's very general is that you can pass any gamedata you want, at any time, so when you say its "undefined" it sounds more like you're not passing something in in the tests, to which the solution is not to delete the tests

louisrli:
the solution is pretty much never to delete the tests

tmansuroglu:
Even when I pass data from parent, test fails . It says gameData is undefined. this is odd.

louisrli:
show me the code

src/components/gamespage/GamesPageImageGrid/index.test.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import '../../../matchMedia';
+import renderer from 'react-test-renderer';
+import GamesPageImageGrid from '.';
+
+it('GamesPageImageGrid renders correctly', () => {
+    const tree = renderer.create(<GamesPageImageGrid />).toJSON();
```
louisrli: i'm not gonna say what's wrong because i want you to figure it out but i want you to look at this line and think about why somtehing might be undefined

and think about whether test code belongs only in test.js files or in GamesPage/index.jsx

tmansuroglu: oh god :D I see it

tmansuroglu: I was about to implement some dependency changes lol

tmansuroglu: PTAL. To keep data same everywhere, I exported data from container and imported it from test. If there is a better way to do it, please let me know.

src/containers/GamesPage/index.jsx
```diff
@@ -2,11 +2,19 @@ import React from "react";
 import GamesPageImageGrid from "../../components/gamespage/GamesPageImageGrid";
 import BottomCurve from "../../components/BottomCurve";
 
+export const gameData = {
```
louisrli: when you are providing test data, you want it to be in the test only. 

as a general rule, we never have test data in production code

the point is not whether your component works on "real data", its whether it works for "any gamedata of valid format i provide it"

basically, move this as MOCK_GAME_DATA into your test.js

and here, you should be providing the real data, not the fake data.

tmansuroglu: PTAL. I also believe that not passing props to test component was the problem with memory game tests. Because invariant function was trying to find length of undefined. That undefined props was being passed from parent.

tmansuroglu: I will create an issue for it

louisrli (approved):

tmansuroglu:
@Ammar-64 ready for merge

tmansuroglu:
@Ammar-64 ready for merge

tmansuroglu:
@Ammar-64 ready for merge

PR #83, opened 2020-11-19
Resources page

YamanSammani:
This PR also connected with Firebase Authorisation. It contains the Resources page with three PDF resources uploaded on firebase.


louisrli (commented):
can you please proofread the "files viewed" and fix any issues you see there

PR #86, opened 2020-11-19
Profile page container

ibrahimaljarad:
small pr for the profile page container


louisrli (approved):

louisrli (changes requested):
is there something else thats gonna be in here later? or else you can just directly use ProfilePage

ssinikoglu (approved):

PR #103, opened 2020-11-21
Fix send button of Contact Form #95

### LINKED ISSUE #95: contact page
removing the tweeter  section and the button, and  fixing the for issue of the forms. adding the spinner  to the sending button

ssinikoglu:
This PR consists of;

-Adding spinner to the send button,
-Create state for spin effect,
-Remove old icon and it's dependencies 

Upcoming PRs' for Contact Page are fixing responsiveness of Contact Form, making adjustments (size,position,responsiveness)  in images at Bottom of the page.
 
I'm looking forward to your reviews.


src/components/ContactUsForm/index.jsx
```diff
@@ -37,6 +40,7 @@ const ContactUsForm = () => {
     }).then(() => {
       setTimeout(() => setIsSubmitted(false), 2000);
       setIsSubmitted(true);
+      setIsLoading(true);
```
majdajroudi: where is isLoading set to false after that?

majdajroudi (approved):

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for reviewing, it stops spinning after post submission message pops out. I'm undecided about where to add change it's state to false again.

ssinikoglu: Do you have any ideas where can I add it?

ssinikoglu: Thanks @majdajroudi for helping me about setting the state to false. I have updated PR, PTAL.

src/components/ContactUsForm/index.jsx
```diff
@@ -35,8 +38,12 @@ const ContactUsForm = () => {
       headers: { "Content-Type": "application/x-www-form-urlencoded" },
       body: encodeFormData({ "form-name": "contact", ...formValue }),
     }).then(() => {
-      setTimeout(() => setIsSubmitted(false), 2000);
+      setTimeout(() => {
```
louisrli: as mentioned you should never use settimeout, i shouldn't have let this past code review so that was my mistake

you are assuming the contact form will take < 2 seconds (you sohuld never make any assumptions about time it takes for something to happen like this)

you want it to do "setisLoading" before you call the fetch, under line 34.

then you want it to setIsLoading to false in the ".then"

you probably want to delete the line about setIsSubmitted(false). because you don't really want the submission msg to go away after the user does it. or use antd.message

ssinikoglu: Thanks for reviewing, I tried following the steps as you have mentioned, PTAL.

src/components/ContactUsForm/index.jsx
```diff
@@ -29,14 +29,14 @@ const ContactUsForm = () => {
...
+    setIsLoading(true);
     fetch("/", {
       method: "POST",
       "data-netlify": "true",
       headers: { "Content-Type": "application/x-www-form-urlencoded" },
       body: encodeFormData({ "form-name": "contact", ...formValue }),
     }).then(() => {
-      setTimeout(() => setIsSubmitted(false), 2000);
-      setIsSubmitted(true);
+      setIsLoading(false);
```
louisrli: you still want to show the postsubmission message. you just want to show it after its submitted. so you still want to keep setIsSubmitted(true), you just dont want to set it back to false after. please give that a try, if you understand the process of what your code is doing then it should make sense :) 

* you fetch
* you wait until fetch finishes
* you change things in the page in the .then()

so think about how you can make the submission message show up

ssinikoglu: It seems, I have misunderstood your first comment.

ssinikoglu: For showing postSubmissionMessage, actually I also tried ant design success message via the [link](https://codesandbox.io/s/oq9d1) that I have found in the documentation.I created success function and pass it to onClick at the button and it worked. The problem, it gave an error when I put translation code instead of a string message. The error was "identifier is needed".  I also had the same problem when I was working to show postSubmissionMessage using a function. Since I couldn't solve the problem, that's why I used setIsSubmitted instead.

`const success = () => {
  message.success('This is a success message');
};`

ssinikoglu: I just want to make an explanation about it, I will do my best for understanding the process and make it work.

louisrli: the success message appears when you call `  message.success('This is a success message');`

so the question is: when do we want the success message to appear?

we want the success message to appear when the fetch finishes. which part of the code is when the fetch finishes?

YamanSammani (approved):
Thanks you Sule. I have no comment rather than what Majed and Louis wrote.

src/components/ContactUsForm/index.jsx
ssinikoglu: `.then()` part the fetch finishes.

ssinikoglu: Post submission message is working right now, PTAL.

ssinikoglu:
There were also problems with catch function and event preventdefault. Thanks to @majdajroudi they are fixed right now.

src/components/ContactUsForm/index.jsx
```diff
@@ -22,41 +23,44 @@ const ContactUsForm = () => {
...
+          content: t("contact.contactUsForm.postSubmissionMessage"),
+          className: "contactUsForm__postSubmissionMessage",
+        });
+        setFormValue({
+          name: "",
+          email: "",
+          message: "",
+        });
+      })
+      .catch((error) => alert(error));
```
louisrli: message.error(error)

src/components/ContactUsForm/index.jsx
```diff
@@ -110,7 +114,9 @@ const ContactUsForm = () => {
           <div className="contactUsForm__submit">
             <button type="submit" className="contactUsForm__submitButton">
               {t("contact.contactUsForm.send")}
-              <RiSendPlaneFill />
+              <span>
+                <SyncOutlined spin={isLoading ? true : false} />
```
louisrli: spin={isLoading}

louisrli (approved):

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks for your review, I updated error function, PTAL.

src/components/ContactUsForm/index.jsx
ssinikoglu: Thanks, I updated the line, PTAL.

ssinikoglu:
Hi @Ammar-64, PR is approved PTL.

PR #99, opened 2020-11-20
passwordReset component

tmansuroglu:
This is separated from #81

PR #110, opened 2020-11-21
Error message component

tmansuroglu:
This separated from #99


louisrli (commented):
proofread

tmansuroglu:
@louisrli extra snapshots are cleared from PR. PTAL

src/components/userauth/ErrorMessage/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ErrorMessage snapshot test 1`] = `null`;
```
louisrli: this test just says null.

add a second test where you pass an error message please

tmansuroglu: I didn't understand what you meant by second test but I passed a test message as prop to the test component. Now snapshot is not null

louisrli: you are allowed to have more than one test in a file. and you should most of the time, especially if there are two cases

src/components/userauth/ErrorMessage/index.test.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import ErrorMessage from ".";
+
```
louisrli, line 4: "it should return null if there is no error"

src/components/userauth/ErrorMessage/__snapshots__/index.test.js.snap
tmansuroglu: So I should be displaying all conditions inside snapshot. I get it.

src/components/userauth/ErrorMessage/index.test.js
tmansuroglu: Returning null doesn't mean something is wrong. Message only displayed when user makes mistake. Most of the time there shouldn't be a message.

tmansuroglu: I feel like im missing your point.

louisrli: your code is correct but i'm aware that returning null doesnt mean something is wrong and that its intended -- thats why we write a test case to express that intention

louisrli (approved):

src/components/userauth/ErrorMessage/index.test.js
louisrli: but in general if you have only 1 test and it all it returns is null its useless for a snapshot. so in this case we definitely want the non-null case covered.

louisrli: for the null case we wouldnt normally use a snapshot test to be honest. we'd just use jest expect(foo).toBeNull

ssinikoglu: I'm also not sure why exactly it is, but null type is defined as object ( I've just found it via MDN [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null) and in the code, message is defined as string. This difference may cause a problem.

tmansuroglu: hmm I see. Thank you Louis

ssinikoglu (approved):

tmansuroglu:
@Ammar-64 can you merge?

PR #111, opened 2020-11-21
registerHelper function

tmansuroglu:
seperated from #99


src/components/userauth/registerHelper.js
```diff
@@ -0,0 +1,12 @@
+export function validateEmail(email) {
+  // regex source https://emailregex.com/
+  const re = /^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/;
+  return re.test(email);
+}
+
+export const createErrorClassName = (error) => {
+  if (!error) {
+    return null;
```
louisrli: you most likely want to return undefined here, otherwise if you do `className={null}` it does `className="null"` which is annoying.

louisrli: className=undefined causes it to be empty in snapshots. you can google this too

louisrli (approved):

YamanSammani (approved):

src/components/userauth/registerHelper.js
tmansuroglu: PTAL

tmansuroglu:
@Ammar-64 can you merge?

PR #125, opened 2020-11-23
Memory game language change modal

tmansuroglu:
Memory game requires page reload to implement language change. I thought I could switch language live with simple code changes but turns out it requires code rework. Instead I created a modal that warns user when language is being changed while game is being played and reloads if user confirms.

Test is not included because I get errors similar to memory game's tests. Ammar was supposed to look at them. I think tomorrow he will fix them. I will update test with him.


src/components/LanguageChangeModal/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <Modal
+      title={t("languageChangeModal.title")}
+      visible={modalVisibility}
+      cancelText={t("languageChangeModal.cancelText")}
+      okText={t("languageChangeModal.okText")}
+      onOk={() => {
+        setModalVisibility(false);
+        // prevents language to change before modal closes
+        setTimeout(() => {
```
ssinikoglu, line 24: Our cases are very different but when I see setTimeout, I was also using setTimeout for post submission message and according to the feedback from Louis, manually setting time for waiting users is not the ideal practice. You may apply [modal](https://ant.design/components/modal/) from ant design.

ssinikoglu (approved):
Code looks clean and easy to understand. I have only left an idea, because I'm not sure it is true for your case.

louisrli (commented):
this looks way too significant of a change to do after code freeze, especially with tests breaking

you can submit this after your demo, but i'm not approving this one, and i don't trust that it's going to work

just don't demo language change when you play the memory game (or do a clever alt tab or something)

louisrli (commented):
also it kinda doesnt make sense to me that it requires a page refresh to change the language of memory game. instead of warning the user the fundamental bug should be fixed anyway

louisrli:
and what sule says about set timeout is true, using settimeout is wrong in
99% of cases

tmansuroglu:
Okay I will come back to this after demo.

louisrli:
anyway, it's an interesting problem; i took a look at your code and can see
why it happens. what you want to do is again have everything by 'image
key'; card state should know nothing about translations. they should be
different layers. this is related to the idea of completely separating the
rendering logic (what to show on the screens, aka translations in this
case) and the game state logic (how does the game work).

it would require a bit of refactoring but it's a good lesson in decoupling,
i recommend giving it a try, to understand code separation like this
(google separation of concerns, decoupling, etc). i'd actually say you're
lucky to have this opportunity to think about this :)

basically, image keys should be the way to look up everything.

game state layer
this could be redux, useReducer, or a class (not sure how to do it with a
class, which would be nice and clean to completely separate the game from
being related to react), i'd probably start by writing useReducer (which is
basically redux)
game state -> based solely on image / card keys. knows nothing about how
it's shown.

rendering layer
renders based on image/card keys present. takes a single object called game
state (this is common in games: google the term "game state") and shows it.
clicking on things dispatches actions. based on the game state, it shows
the translation

tmansuroglu:
I will do it. Thank you @louisrli .

PR #144, opened 2020-11-23
Debug resources page

YamanSammani:
In this Pr, I fixed the mouse hover and fixed the resources translation.


ibrahimaljarad (approved):

src/containers/ResourcesPage/index.jsx
```diff
@@ -6,39 +6,34 @@ import resourcesForEducation from "../../images/resourcesForEducation.png";
...
-  },
-];
-
 const ResourcesPage = () => {
   const [t] = useTranslation();
-
+  const initResources = [
+    {
+      id: 1,
+      title: t("resourcesPage.title1"),
```
louisrli: use an array, almost never call something title1 title2

louisrli: an array in the translations. work with tarkan if needed

YamanSammani: @louisrli , for the naming, I will update them. but regarding the array, could you kindly elaborate more. I don't see the point of using an array in here.

YamanSammani: PTAL

src/containers/ResourcesPage/index.jsx
```diff
@@ -6,39 +6,34 @@ import resourcesForEducation from "../../images/resourcesForEducation.png";
...
-  },
-];
-
 const ResourcesPage = () => {
   const [t] = useTranslation();
-
+  const initResources = [
+    {
+      id: 1,
+      title: t("resourcesPage.resourceOneTitle"),
```
louisrli: resourceOne is not better than title1

they also have the same structure, thats why an array makes sense, either that or meaningful names

your two options are to make the code either look like this:

t("resourcesPage.resources[0].title")
t("resourcesPage.resources[0].text")
(or something like this, i cant remember the syntax for arrays)

or like this:
t("resourcesPage.resources.climateChangeHandbook.text") (or some other meaningful name)

both of these are cleaner, the second provides more meaning if the descriptions actually have to do with the resource themselves. which it does.

imagine your teammate needs to change this code, and they need to change the order of the arrays. and they need to decide ,"hey should i use resourceOneText or resourceTwoText for climate change handbook.pdf?" 

* more readable: t("resourcesPage.resources.climateChangeHandbook.title")
* not readable: (".resourceOneText")

YamanSammani: in this way, it's much much much clearer :D, Thanks. PTAL.

src/components/Resource/index.css
```diff
@@ -45,3 +47,7 @@
   padding: 6px 10px;
   border-radius: 7px;
 }
+.resource__body__content__footer button:hover {
```
louisrli, line 51: newline above

src/containers/ResourcesPage/__snapshots__/index.test.js.snap
```diff
@@ -111,12 +111,12 @@ exports[`ResourcesPage renders correctly 1`] = `
...
           className="resource__body__content__title"
         >
-          Who Causes Climate changes - TROCAIRE
+          resourcesPage.resource.whoCausesClimateChangesHandbook.title
         </div>
         <div
           className="resource__body__content__description"
         >
-          Children explore who contribute to climate change by looking at activities related to climate change using climate change bingo sheet, creating clouds that list activates causing gas emissions and represent the world population by dividing themself.
+          resourcesPage.resource.whoCausesClimateChangesHandbook.titleext
```
louisrli: titleext

src/containers/ResourcesPage/index.jsx
```diff
@@ -6,39 +6,38 @@ import resourcesForEducation from "../../images/resourcesForEducation.png";
...
+      body: t("resourcesPage.resources.educationOnClimateChangeHandbook.text"),
+      pic: resourcesForEducation,
+      url:
+        "gs://capstone-istanbul-juniper.appspot.com/ResourcesPage/Resources list_Climate Change.pdf",
+    },
+    {
+      id: 3,
+      title: t("resourcesPage.resource.whoCausesClimateChangesHandbook.title"),
+      body: t(
+        "resourcesPage.resource.whoCausesClimateChangesHandbook.titleext"
```
louisrli: titleext

YamanSammani: I have a PhD in Typo :D, PTAL.

src/components/Resource/index.css
YamanSammani: PTAL

louisrli:
you should also proofread your work :)

src/locales/en.json
```diff
@@ -277,8 +277,22 @@
...
+        "title": "Our Climate Change",
+        "text": "This book is a child friendly handbook meant to explain the concepts of climate change in an easily understandable manner and language. It aims to link behavioural choices to the changes observed in our climate."
+      },
+      "educationOnClimateChangeHandbook": {
+        "title": "Resources for Education on Climate Change - UNESCO",
+        "text": "This publication focuses on the challenges, opportunities and good practices of climate change."
+      },
+      "whoCausesClimateChangesHandbook": {
+        "title": "Who Causes Climate changes - TROCAIRE",
+        "test": "Children explore who contribute to climate change by looking at activities related to climate change using climate change bingo sheet"
```
louisrli: fix, and please test your code and translations

louisrli: and proofread everything before sending it again

YamanSammani:
@louisrli , PTAL. I hope it's clean now 😅.

louisrli (approved):

